PR #2, opened 2012-12-06
Fix control button size rendering on mobile

RhubarbRhubarb:
Fixes the control button size as rendered on mobile web browsers.

The locate button is now the same size as the zoom in / out buttons.
Before the locate button was much smaller, and much harder to push with fat fingers.

Tested on Fennec (mobile firefox 15.0) and the default web-kit based browser on the Nokia n9 Meego phone.

### DIFF
--- a/src/L.Control.Locate.css
+++ b/src/L.Control.Locate.css
@@ -34,3 +34,8 @@
 .leaflet-control-locate.active a {
     background-image: url(images/locate_active.png);
 }
+
+.leaflet-touch .leaflet-control-locate a {
+    width: 27px;
+    height: 27px;
+}


domoritz:
@RhubarbRhubarb Thanks for the pull request. However, are you sure that 27px is the right size? The original leaflet controls have 30px (see https://github.com/CloudMade/Leaflet/blob/master/dist/leaflet.css#L238).

RhubarbRhubarb:
I'm not sure domoritz,

I was working off the stable release, where the controls are 27px (see https://github.com/CloudMade/Leaflet/blob/stable/dist/leaflet.css#L140).

I must say using 30px controls on my phone is much better than the 27px ones :smile:

domoritz:
@RhubarbRhubarb I should have paid more attention to the right version ;-)

I'll keep the 27px for now and change the size to 30px once the stable version has also changed.

PR #10, opened 2013-02-06
Location errors should handle via callbacks

omjokine:
Instead just calling alert(err.message) handle error and pass it to callback function.


domoritz:
@omjokine Can you add this to the documentation as well?

PR #11, opened 2013-03-12
Configurable title attribute

comunidadfusa:
I added an option attribute which can be configured in order to change the title of the locate icon.


domoritz:
Thanks for the pull request. One last thing before I merge this: Why did you change the indention? Is there a certain code style that one should use or is this just your personal preference? I would happily accept a proper code style but I don't want to change it for no reason.

comunidadfusa:
My mistake! :-P I replaced the whole code instead of adding just the new portions of code. I'll be re-pulling a new version without indention changes, ok?

domoritz:
Great.

comunidadfusa:
I created a new pull request.

PR #18, opened 2013-04-25
Update button style for touch devices (tested on android webview)

bpavot:
Here is the updated version using leaflet-bar-part.
I also reused the leaflet-control-zoom style to maximize reuse (Just pull the leaflet-bar-part-single class if you don't want the zoom class)


domoritz:
@bpavot Thanks, merged it.

PR #22, opened 2013-06-10
Add option to cancel following on map drag

radicalbiscuit:
I've added an option, `cancelFollow`, to cancel following when the map is dragged if `follow` is set to true.


domoritz:
@radicalbiscuit This is a super useful feature. However, we need visibility of system status to show users what's happening. Could you add another state to the control to indicate that the map follows the user. So we would have `requesting`, `active` and `following`. I'd suggest we have an indicator like the active one (with the blue dot) but with an orange dot. I can create the image, if you are willing to write the code that adds and removes the class.

domoritz:
Oh and can we rename the option to `stopFollowingOnDrag` to make it more clear what it does?

radicalbiscuit:
No problems. I'd like to take a shot at the images too, if you don't mind. Give me a little bit and I should be done with the updates.

radicalbiscuit:
Okay, take a look and let me know if I went overboard.

domoritz:
Hey, thanks for your work. I fixed a few minor things and recreated the image from the svg.

PR #26, opened 2013-06-30
Respect map boundaries

### LINKED ISSUE #25: Consider the map boundaries
There is a possibility to define map boundaries in Leaflet (`maxBounds` option), such that the user cannot pan outside some defined area. It would be interesting to catch the cases when the location is outside the map boundaries, and offer the developer an option to not set the view to this location.

jeromecaffaro:
When the map has defined boundaries (maxBounds), do not move
to the location if it is outside the boundaries.
Added "onLocationOutsideMapBounds" option to trigger a callback
and customize the behaviour. By default move is aborted and
message from newly defined "options.strings.outsideMapBoundsMsg"
is displayed. Fixes #25

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -32,10 +32,12 @@ L.Control.Locate = L.Control.extend({
         onLocationError: function(err) {
             alert(err.message);
         },
+        onLocationOutsideMapBounds: null,
         setView: true, // automatically sets the map view to the user's location
         strings: {
             title: "Show me where I am",
-            popup: "You are within {distance} {unit} from this point"
+            popup: "You are within {distance} {unit} from this point",
+            outsideMapBoundsMsg: "You seem located outside the boundaries of the map"
         },
         locateOptions: {}
     },
@@ -75,7 +77,8 @@ L.Control.Locate = L.Control.extend({
             .on(link, 'click', L.DomEvent.stopPropagation)
             .on(link, 'click', L.DomEvent.preventDefault)
             .on(link, 'click', function() {
-                if (self._active && (map.getBounds().contains(self._event.latlng) || !self.options.setView)) {
+                if (self._active && (map.getBounds().contains(self._event.latlng) || !self.options.setView ||
+                    (map.options.maxBounds && !map.options.maxBounds.contains(self._event.latlng)))) {
                     stopLocate();
                 } else {
                     if (self.options.setView) {
@@ -131,9 +134,15 @@ L.Control.Locate = L.Control.extend({
 
         var visualizeLocation = function() {
             var radius = self._event.accuracy / 2;
-
             if (self._locateOnNextLocationFound) {
-                map.fitBounds(self._event.bounds);
+                if (map.options.maxBounds &&
+                    !map.options.maxBounds.contains(self._event.latlng) &&
+                    !self.options.onLocationOutsideMapBounds(self._event)) {
+                    // outside map boundaries
+                    self._following = false;
+                } else {
+                    map.fitBounds(self._event.bounds);
+                }
                 self._locateOnNextLocationFound = false;
             }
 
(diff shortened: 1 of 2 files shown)


domoritz:
@jeromecaffaro Can you set `options.onLocationOutsideMapBounds` to the function you defined for `onLocationOutsideMapBounds` instead of setting it no null?

jeromecaffaro:
@domoritz Done.

I had to move it as an anonymous function, and pass the message string as parameter, to overcome scope issues.
Any comment on a better way to do it is welcome :-)

src/L.Control.Locate.js
```diff
@@ -131,9 +137,16 @@ L.Control.Locate = L.Control.extend({
 
         var visualizeLocation = function() {
             var radius = self._event.accuracy / 2;
-
             if (self._locateOnNextLocationFound) {
-                map.fitBounds(self._event.bounds);
+                if (map.options.maxBounds &&
```
domoritz: Could you move this whole thing into an internal function and then remove the comment (if you choose a good name, it will not be necessary any more). Thanks!

src/L.Control.Locate.js
```diff
@@ -75,7 +80,8 @@ L.Control.Locate = L.Control.extend({
             .on(link, 'click', L.DomEvent.stopPropagation)
             .on(link, 'click', L.DomEvent.preventDefault)
             .on(link, 'click', function() {
-                if (self._active && (map.getBounds().contains(self._event.latlng) || !self.options.setView)) {
+                if (self._active && (map.getBounds().contains(self._event.latlng) || !self.options.setView ||
```
domoritz, line 82: Same here, it's a hell of a condition that should be composed of semantic function calls that can also be reused in lines 141 to 144.

src/L.Control.Locate.js
domoritz: Also, why is `!self.options.onLocationOutsideMapBounds(self._event, self.options.strings.outsideMapBoundsMsg)` in the condition of the if? It's very difficult to understand when it's called.

domoritz:
@jeromecaffaro Thanks! However, I'm still not really happy (which has nothing to do with your pull code which is good so far). It's just the fact that the message is only used in the function  and it seems that the message is only used in `onLocationOutsideMapBounds` and other implementation might not even need it. I think I made up my mind and prefer passing `self` and call it `context`. This way other things can be accessed from other implementation that people might come up with. 

Sorry for having so many wishes and comments. I like your code and the improvement for the control is great. However, I would like to have an API that does not need to change soon and an implementation that others can understand and improve.

Again, thanks for your pull request!

domoritz:
> it might be decided that it is not needed to override the behavior, such that the function would not return any value, and the map would never be re-centered on the location when off the bounds. It would further simplify the code around lines 146-153

I think I prefer this option. If someone set's map bounds, they probably have a reason for that. This could be that they don't have tiles for the area for example.

src/L.Control.Locate.js
```diff
@@ -129,11 +134,20 @@ L.Control.Locate = L.Control.extend({
...
+        }
+
         var visualizeLocation = function() {
             var radius = self._event.accuracy / 2;
-
             if (self._locateOnNextLocationFound) {
-                map.fitBounds(self._event.bounds);
+                if (isOutsideMapBounds()) {
+                    self.options.onLocationOutsideMapBounds(self);
+                    self._following = false;
```
domoritz, line 147: is there a reason not to call `stopFollowing()`?

domoritz:
@jeromecaffaro I made one last comment and will merge your pull request when you've merged the main branch into your brach so that it can be merged automatically. 

Thank you for your work. I should probably add a list of contributors somewhere...

src/L.Control.Locate.js
jeromecaffaro, line 147: As such we would enter in an infinite recursion, isn't it? `visualizeLocation()->stopFollowing()->visualizeLocation()->...`

domoritz, line 147: True... In that case just add a comment that says that, merge master in ad I'll merge it in.

jeromecaffaro:
I have merged your main branch into mine. I will let you rebase and squash as appropriate once you are ok to merge.

domoritz:
@jeromecaffaro Awesome. It's merged.

PR #30, opened 2013-07-03
fixing blue image offset

gabrielflorit:
The blue circle was offset to the right by 1 pixel.

Before:
screen shot 2013-07-03 at 12 32 31 pm

After:
screen shot 2013-07-03 at 12 32 41 pm


domoritz:
@gabrielflorit I already had a local fix for that when you submitted your pull request. Thanks for your effort!

gabrielflorit:
Nice - thanks for the control!

PR #31, opened 2013-07-03
remove unnecessary CSS declarations

gabrielflorit:
Unless I'm mistaken, you don't need to class it with `leaflet-control-zoom`.

Also, all its corners will be rounded by the leaflet CSS, so no need to round them yourself.


domoritz:
Looks good. Thanks!

Issue #32, opened 2013-07-04
Documentation should say "popup" and not "popupText"


fnicollet:
Hello,
In the docs, it says you can change "popupText" but it looks like the property is actually "popup".

This produce a bug when the L.Template is being applied

Fabien

domoritz:
Fixed with #33.

PR #33, opened 2013-07-04
String should be popup, not popupText

fnicollet:
see https://github.com/domoritz/leaflet-locatecontrol/issues/32


domoritz:
Good catch. I think I changed it at some point because I added a group `strings` and the `...text` seemed unnecessary.

PR #42, opened 2013-11-02
Add maxZoom option.

davidkeen:
Add a maxZoom option to locateOptions.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -170,6 +170,9 @@ L.Control.Locate = L.Control.extend({
                     self.options.onLocationOutsideMapBounds(self);
                 } else {
                     map.fitBounds(self._event.bounds);
+                    if (self.options.locateOptions.maxZoom) {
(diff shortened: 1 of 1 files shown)


domoritz:
Heya. Thanks for your pull request. I'm sorry to reject this as it is but the behaviour is not correct. `maxZoom` is the maximum zoom level, not the actual zoom level. That's why I use `map.fitBounds(self._event.bounds);` which chooses the correct level to show the whole circle. 

If you can fix this and send a new pull request, I'll merge.

davidkeen:
Ah ok. The functionality works as expected for me (well at least as I would like it to so maybe we are talking about different things. :smiley:). It actually calls map.setZoom with the maxZoom value so perhaps maxZoom is not a good name for it. Perhaps it should just be 'zoom'.

For example: `L.control.locate({locateOptions: {zoom: 10}}).addTo(map);`

domoritz:
Yeah, but I guess that's not what you want in the case where the circle is large.

PR #44, opened 2013-11-18
Add event firings to startFollowing and stopFollowing

### LINKED ISSUE #43: Feature: stopFollowingOnEvents
I introduced the `stopFollowingOnDrag` functionality in #22. This has proven to be a very handy feature for our uses, but it has recently come to our attention that one might want following to cancel on any number of events, `dragstart` (or any of the other drage events) being one.

In our local copy, I've written functionality for `stopFollowingOnEvents` into the library. Upon initialization, you declare something like the following in the init options:

```
stopFollowingOnEvents: [
    'dragstart',
    'mousedown',
    'customEvent'
]
```

Inside the library, each item from the provided array is taken and the following-cancellation event handler is bound to it.

This works well. However, we have further customization inserted (awareness of custom events we added into an old version of Leaflet), so I'll have to rework it for a patch, which I'm happy to do.

So, this issue exists in effect to find out if this would be a welcome feature, and if so, what should be done with `stopFollowingOnDrag`. If desired, `stopFollowingOnDrag` could simply become an alias to `stopFollowingOnEvents: ['dragstart']` to preserve backward compatibility.

Anyhow, lemme know. As I said, I'll be happy to make a pull request, but I wanted to see if it'd be welcome, first.

radicalbiscuit:
Resolves #43.

This PR adds Leaflet map event firings to `startFollowing` and `stopFollowing`, thereby allowing developers to take action based on following status.

It also deprecates the `stopFollowingOnDrag` option in favor of binding a `stopFollowing` call to arbitrary events.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -9,7 +9,7 @@ L.Control.Locate = L.Control.extend({
         position: 'topleft',
         drawCircle: true,
         follow: false,  // follow with zoom and pan the user's location
-        stopFollowingOnDrag: false, // if follow is true, stop following when map is dragged
+        stopFollowingOnDrag: false, // if follow is true, stop following when map is dragged (deprecated)
         // range circle
         circleStyle: {
             color: '#136AEC',
@@ -140,6 +140,7 @@ L.Control.Locate = L.Control.extend({
         };
 
         var startFollowing = function() {
+            map.fire('startfollowing')
             self._following = true;
             if (self.options.stopFollowingOnDrag) {
                 map.on('dragstart', stopFollowing);
@@ -147,6 +148,7 @@ L.Control.Locate = L.Control.extend({
         };
 
         var stopFollowing = function() {
+            map.fire('stopfollowing');
             self._following = false;
             if (self.options.stopFollowingOnDrag) {
                 map.off('dragstart', stopFollowing);
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ L.control.locate({
 	position: 'topleft',  // set the location of the control
     drawCircle: true,  // controls whether a circle is drawn that shows the uncertainty about the location
     follow: false,  // follow the location if `watch` and `setView` are set to true in locateOptions
-    stopFollowingOnDrag: false, // stop following when the map is dragged if `follow` is set to true
+    stopFollowingOnDrag: false, // stop following when the map is dragged if `follow` is set to true (deprecated)
     circleStyle: {},  // change the style of the circle around the user's location
     markerStyle: {},
     followCircleStyle: {},  // set difference for the style of the circle around the user's location while following


src/L.Control.Locate.js
```diff
@@ -9,7 +9,8 @@ L.Control.Locate = L.Control.extend({
         position: 'topleft',
         drawCircle: true,
         follow: false,  // follow with zoom and pan the user's location
-        stopFollowingOnDrag: false, // if follow is true, stop following when map is dragged
+        stopFollowingOnDrag: false, // if follow is true, stop following when map is dragged (depricated)
```
domoritz: deprecated

src/L.Control.Locate.js
```diff
@@ -189,6 +219,9 @@ L.Control.Locate = L.Control.extend({
                         .addTo(self._layer);
                 } else {
                     self._circle.setLatLng(self._event.latlng).setRadius(radius);
+                    for (styleOption in style) {
```
domoritz: What is style? This doesn't seem to belong here.

src/L.Control.Locate.js
```diff
@@ -215,9 +248,12 @@ L.Control.Locate = L.Control.extend({
                     .bindPopup(L.Util.template(t, {distance: distance, unit: unit}))
                     .addTo(self._layer);
             } else {
-                self._circleMarker.setLatLng(self._event.latlng)
-                    .bindPopup(L.Util.template(t, {distance: distance, unit: unit}))
+                self._circleMarker.setLatLng(self._event.latlng);
+                self._circleMarker.bindPopup(L.Util.template(t, {distance: distance, unit: unit}))
                     ._popup.setLatLng(self._event.latlng);
+                for (mStyle in m) {
```
domoritz: Same here.

src/L.Control.Locate.js
domoritz: Or is this "In addition, a regression involving the circle markers was fixed. A previous commit changed circle markers to use a persistent object, but made no provision for refreshing the style, thereby excluding cases where the circle styles are different depending on the following status."?

radicalbiscuit: Yes, that's the reason for this bit. Should've raised a new issue (and I still can), but thought I'd take care of it when I saw it.

src/L.Control.Locate.js
radicalbiscuit: Yes, same thing. Dealing with the circle style regression.

src/L.Control.Locate.js
domoritz: Cool. can we make this a separate pr? Also, there is a `var` missing. It should be: `for (var styleOption in style) {`.

radicalbiscuit:
Something else to consider: I foresee a common use of this would be to stop following on zoom. However, if I bind to `zoomstart`, this is fired by `fitBounds()` which is called in `visualizeLocation`.

This is where our custom build of Leaflet has come into play. We added several more events to be fired for the specific zoom events (`scrollwheelzoomstart`, `scrollwheelzoomstart`, etc.) so we could differentiate between the zoom initiated by `fitBounds()` and user-initiated zooms.

However, since this might be a very common feature, might we want to provide for the possibility by checking if `zoomstart` or `zoomend` is in `stopFollowingOnEvents` and if so, unbinding them before executing `fitBounds` and rebinding afterward?

I wouldn't be inclined to do it this way, but I think the chances of having a pull request in Leaflet to add events is even less likely.

src/L.Control.Locate.js
radicalbiscuit: Right you are. It can be done.

domoritz:
Looks good! One thing I don't understand is what `_currentStopFollowingOnEvents` is for. Why don't we just enable the events on initialization? Is this so that you can bind and unbind events? 

You seem to have thought about this a lot but I struggle understanding the purpose of the methods and data structures that you added.

src/L.Control.Locate.js
```diff
@@ -162,10 +184,18 @@ L.Control.Locate = L.Control.extend({
...
+            }
 
             var radius = self._event.accuracy;
+
+            if (self._currentStopFollowingOnEvents.length &&
+                    (self._currentStopFollowingOnEvents != self.options.stopFollowingOnEvents)) {
+                _followingEventsOff();
+                map.on(self.options.stopFollowingOnEvents.join(' '), startFollowing);
+            }
+
```
domoritz: I don't understand this. Why is this necessary?

radicalbiscuit:
`_currentStopFollowingOnEvents` is instituted so `options.stopFollowingOnEvents` can be changed dynamically. I don't actually use it this way, but I wanted to provide for the possibility. I will concede that this is not in the spirit of the rest of the library, but I wish it was. For instance, there's currently not an easy way to get `L.Control.Locate` to stop following from outside the library (my current hack is to set `_following` to `false` from outside, but I'd rather be able to call `stopLocate()`).

Removing it would simplify the implementation of `stopFollowingOnEvents` and keeping it would allow for a feature that may not ever be used. I'll leave it up to you. :)

src/L.Control.Locate.js
radicalbiscuit: For situations in which `self.options.stopFollowingOnEvents` has changed.

domoritz:
What do you think about the following:

``` js
var lc = L.control.locate({
    follow: true
}).addTo(map);

map.on('dragstart', lc.stopFollowing);
```

The only change that is needed for this is to make `stopFollowing` available to the outside world. Does that allow you to do all the things that you want?

`stopLocate` can already be called from the outside. See https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/src/L.Control.Locate.js#L274

radicalbiscuit:
That's good news. I wasn't up to date since we were using our out-of-date fork. Very good.

That would do the trick. That would invalidate this pull request so far (which is quite alright) and bring up the question of what to do with `stopFollowingOnDrag`. Still in line for deprecation probably, but would there be any further changes needed (prior to the end of the deprecation period)?

domoritz:
Yeah, I'm sorry that I haven't though of this earlier. I just pushed the change in 08516c96b97e2b791cb5412406742b65f9152535which makes `stopFollowing` available to the outside world. 

The problem with this is that it doesn't unbind the event when `stopFollowing` is called. Do you know how to fire custom events like `stopFollowing`?

Also, your changes to update the style seem useful!

radicalbiscuit:
No problems. I think this is an even better way to deal with it. It may be a little less apparent to someone picking up the library as it won't appear as an option in the easily-referenced options list, but it keeps the library simpler and the usages more liberated.

As far as firing events, that's easy enough with `dispatchEvent` or jQuery's `trigger`. It may be a good idea to simply bind `stopfollowing` (all lowercase to keep with the convention, or at least the status quo) to `stopFollowing()` within the library and allow users to fire the event whenever needed.

I'll give some thought to it over night.

---

After re-reading your previous comment, I'm pretty sure I misunderstood something and my response may not make sense. It's been a long day, and like I said, I'll be better able to share my thoughts tomorrow. :D

domoritz:
Sounds great. If you could send a pr, that would be great. I'd like to avoid jQuery, and use only Leaflet or pure js.

radicalbiscuit:
As the end purpose is the same, I've merely recommitted and altered the title and description of this PR.

Leaflet already had a built-in firing framework for just this sort of thing. Firing an event for both `startFollowing` and `stopFollowing` will allow devs to unbind and bind a series of actions to those methods.

The events fired are `startfollowing` and `stopfollowing`.

domoritz:
Cool. merging now. I'll only add a note to the readme on how to work with the new events.

PR #45, opened 2013-11-18
Circle Marker Fixes

radicalbiscuit:
This branch resolves the inability of circle markers to change style based on the `_following` status. It also moves the `L.circle.setLatLng` calls to the ends of relevant method chains (as this method does not return the object like the other methods do).

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -189,6 +189,9 @@ L.Control.Locate = L.Control.extend({
                         .addTo(self._layer);
                 } else {
                     self._circle.setLatLng(self._event.latlng).setRadius(radius);
+                    for (var styleOption in style) {
+                        self._circle.options[styleOption] = style[styleOption];
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


domoritz:
It does return the circle. See http://leafletjs.com/reference.html#circlemarker-setlatlng.

radicalbiscuit:
I promise I'm not trying to waste your time.

I had run into a bug earlier in the day caused by this, checked the Leaflet source, and saw that it wasn't returning `this`. Probably another case of an outdated Leaflet library. I can't track it down. I'll try and fix my fix.

radicalbiscuit:
There, should be good now.

domoritz:
Yeah, old versions are a problem. I stopped looking at all the old versions. 

Thanks. Merging now.

radicalbiscuit:
[I found it.](https://github.com/Leaflet/Leaflet/blob/master/src/layer/vector/CircleMarker.js#L25)

This should probably be a Leaflet issue (which I'll be happy to create) since it's not following its own API, but this exists in Leaflet's `master` branch and the 0.7 release from today.

Looking at the history, this looks like a regression that's existed for 4 months since `setLatLng` was overridden in `L.CircleMarker`.

So, I guess I'll submit a pull request to Leaflet. But I've found that it can take a long time to have a PR addressed over there.

domoritz:
Especially if there is a new release.

radicalbiscuit:
In a cruel twist of fate, **you** seem to be the impetus for this bug by reporting [this one](https://github.com/Leaflet/Leaflet/issues/1921), which resulted in the offending commit. :smile:

domoritz:
Haha, great. I feel innocent, though ;-)

PR #46, opened 2013-12-10
Style tweaks for Leaflet 0.7

bmcbride:
This update brings the locate control in line with 0.7.1, which introduced a more subtle border around mobile controls. See https://github.com/Leaflet/Leaflet/issues/1868 and https://github.com/Leaflet/Leaflet/commit/ea6095f259a1d3ba950526e446d4f8d11334abae.


domoritz:
Cool, could you give me a short description and use the same white space character? Thanks!

PR #55, opened 2014-02-18
Call stop following on stopLocate

uintaam:
stopfollowing was not triggered when clicking the location icon when location is active.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -256,7 +256,7 @@ L.Control.Locate = L.Control.extend({
         var stopLocate = function() {
             map.stopLocate();
             map.off('dragstart', stopFollowing);
-            if (self.options.follow) {
+            if (self.options.follow && self._following) {
                 stopFollowing();
             }
 
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -256,6 +256,9 @@ L.Control.Locate = L.Control.extend({
         var stopLocate = function() {
             map.stopLocate();
             map.off('dragstart', stopFollowing);
+            if (self.options.follow) {
+                stopFollowing();
+            }
 
             L.DomUtil.removeClass(self._container, "requesting");
             L.DomUtil.removeClass(self._container, "active");


domoritz:
Could you describe a case where this is a problem? The call to `map.stopLocate()` should disable locating so that we don't need to disable following.

uintaam:
For example I'm using this in some code and setting a value isFollowing true or false when following is disabled.  Without the change I've added, stopfollowing is only fired when dragstart is called and not if I click the locate button, to stop locating, after locating has started.

```
var isFollowing = false;
self.map.on('startfollowing', function() {
    isFollowing = true;
}).on('stopfollowing', function() {
    isFollowing = false;
});
```

Edited: to remove code not relevant to this issue, and make it more understandable.

domoritz:
I'm sorry but I still don't understand. You don't want `stopfollowing` to be fired when someone clicks on the button because if you click it again, you wouldn't follow the location any more. 

I am probably missing something so let me know whether I am making sense.

uintaam:
No worries, maybe I'm missing something.  Without this change, here's what I see.
1.) Click Button -> startfollowing fired
2.) Move map manually -> stopfollowing fired
3.) Click button again -> startfollowing fired
4.) Click button again -> stopfollowing NOT fired
5.) Move map manually -> stopfollowing NOT fired
6.) Click button again -> startfollowing fired

It seems like stop following should be called in 4 or and baring that at least step 5. but it's not fired.   

The change I made (just modified slightly) makes the above play out like this.
1.) Click Button -> startfollowing fired
2.) Move map manually -> stopfollowing fired
3.) Click button again -> startfollowing fired
4.) Click button again -> stopfollowing fired
5.) Move map manually -> stopfollowing NOT fired
6.) Click button again -> startfollowing fired

Maybe I'm just not thinking about it right, but if we turn location off via the button, I assumed it would also stop following your location, since it would never update anyway.

domoritz:
Ahh, sorry for not being smart enough to understand the problem in the first place. It makes sense to me now. Thank you for the pr and the explanations!

PR #59, opened 2014-02-27
Add retina icons

tmcw:


domoritz:
Thanks for updating the [CDN](https://www.mapbox.com/mapbox.js/example/v1.0.0/leaflet-locatecontrol/). Is there  way to automatically also add 0.25.0 with the vector icons?

tmcw:
I just added v0.25.0 - mapbox.js plugins are added [like so](https://github.com/mapbox/mapbox.js-plugins/commit/4123bd832f73f5fbe1c010aa863aa642c2f5aceb) - a fairly manual process since the structure & management of plugins differs so much.

PR #61, opened 2014-03-05
Option to use the current zoom level when map is refreshed after a geolocation

gitare:
use
locateOptions: {
           keepCurrentZoomLevel: true,
}
or as before :
locateOptions: {
           maxZoom: xx,
}


src/L.Control.Locate.js
```diff
@@ -54,7 +54,7 @@ L.Control.Locate = L.Control.extend({
             outsideMapBoundsMsg: "You seem located outside the boundaries of the map"
         },
         locateOptions: {
-            maxZoom: Infinity,
```
domoritz, line 57: Why did you remove maxZoom?

src/L.Control.Locate.js
```diff
@@ -54,7 +54,7 @@ L.Control.Locate = L.Control.extend({
...
+            keepCurrentZoomLevel: true,
```
domoritz: This option should not be in `locateOptions` but rather one level up.

src/L.Control.Locate.js
gitare, line 57: I have replaced it by keepCurrentZoomLevel: true. 
My suggestion is to use keepCurrentZoomLevel: true as a default feature (there is then no need to set maxZoom), and otherwise use maxZoom if you want to force a zoom level.

Maybe it would be more clear to use maxZoom and to set a value like maxZoom: 'keepCurrent'? (but then it is not consistent with your second comment below).

I am new to this project, so sorry if I don't understand quite well how it works and why it was designed in a certain way.

domoritz, line 57: I understand that you are new to the project and I have strong options about some things ;-) 

`maxZoom: Infinity` is only used as a default and should stay here. Just move the option one level up and it will be consistent with the current API. Also, please add a documentation to the readme?

gitare:
Here it is:
- keepCurrentZoomLevel at the `options` level
- readme.md updated

domoritz:
Thanks! Merged.

PR #71, opened 2014-06-12
New remainActive option and new onClick hook.

thirstyfish:
I've added two new options that might be useful to others as well:

| Option | Type | Description |
| :-- | :-- | :-- |
| `remainActive` | _boolean_ | Should the locate control remain active on click even if the user's location is in view |
| `onClick` | _function_ | Event hook, called when the user clicks on the locate control |

My use case that I needed this functionality for is the following:

I want users to be able to locate themselves on the map,and also be able to adjust the position of the marker (but not the circle) after they have been located. If the locate control button is clicked again, it moves the marker back to its' original position (hence the `onClick` hook) and does not deactivate the control even if it's in view (hence the `remainActive` option).

This use case is illustrated in this jsfiddle, where I linked in my fork of the leaflet-locatecontrol:
http://jsfiddle.net/LhGEW/

Cheers

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -38,6 +38,9 @@ L.Control.Locate = L.Control.extend({
         iconLoading: 'icon-spinner animate-spin',
         circlePadding: [0, 0],
         metric: true,
+        onClick: function(control){
+            // this event is called when the user clicks on the locate control
+        },
         onLocationError: function(err) {
             // this event is called in case of any location error
             // that is not a time out error.
@@ -49,6 +52,9 @@ L.Control.Locate = L.Control.extend({
             alert(context.options.strings.outsideMapBoundsMsg);
         },
         setView: true, // automatically sets the map view to the user's location
+        // locate control remains active on click even if the user's
+        // location is in view.
+        remainActive: true,
         // keep the current map zoom level when displaying the user's location. (if 'false', use maxZoom)
         keepCurrentZoomLevel: false,
         strings: {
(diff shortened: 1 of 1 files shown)


domoritz:
Great. Can you make the `onClick` and even that is triggered rather than an event hook?

domoritz:
Related to #72. @thirstyfish I believe you could get the click event without any modifications.

domoritz:
Actually, I was thinking about this and don't see why this should be integrated into the main repo. To be honest, I am not convinced that there are other use cases than the one you described. Do you have other ideas? What are alternatives to your approach that might generalize better?

thirstyfish:
Sure, no worries.
What do you reckon the best approach would be to listen to the click event ?
`$("a.icon-location").on("click", function(){ ... })` is not too nice.

Finally jsfiddle is back online, so you can have a look at my use case, but you're right that it might be a rare requirement. 
I guess the generic use case would be if the developer doesn't want to allow the user to turn off the location control (I could see this being a requirement).

gmccrackin:
I think you might want to use the getContainer method on the control.

``` js
var lc = L.control.locate().addTo(map);

$(lc.getContainer()).children('a').on('click', function (e) { ... });
```

domoritz:
I don't want to include the click handler code and think @gmccrackin's solution is better than an explicit handler. I like the `remainActive` option though and will merge this pr or #73 depending on which pr is adapted faster. This pr is missing docs in the readme and I personally prefer the one line implementation from #73.

domoritz:
Duplicate of #73

PR #75, opened 2014-07-03
Fix out-of-bounds alert and only show it once

stuporglue:
This pull request: 

1) Fixes the out-of-bounds alert (context is never defined)
2) Prevents repeated out-of-bounds notifications
3) Adds some missing semicolons


kaio328:
"Fix out-of-bounds alert and only show it once",

its definetly not fixed :(.  when the users location is "notInBounds" every following location attempt gets canceled and results in the errormsg.

domoritz:
@kaio328 Fixed in 2322cb4

kaio328:
u saved my saturday 👍 :)  ! BIG THANKS !

also solved my spinner problem!

PR #77, opened 2014-07-10
Allow the popup to be disabled with an option.

MobileSam:
In addition to be able to control the text that shows in the popup, we should be able to disable it completely. This pull request add this option.


domoritz:
can you add docs to the readme?

MobileSam:
I updated the documentation with the new parameter.

domoritz:
The popup string is always defined. No need to check it in the condition.

PR #86, opened 2014-08-18
Update README

marekkalnik:
Add default values for locateOptions


domoritz:
Thanks. Not really relevant though for most users, imho.

PR #109, opened 2014-11-23
Refactoring - second try

marekkalnik:


domoritz:
Awesome, here are a few small things that need to be done before I merge:
- [ ] When following is enabled (as in the demo), and I do the following: click the (scrolls, icon is orange), pan the map so that the circle is outside the map view (icon is blue now) and then click the control again (to get the circle back into the view), the icon should now be orange and following enabled but it's not.
- [ ] Using the same procedure, I get a bunch of errors
  screen shot 2014-11-24 at 12 42 58
- [ ] The fact that there are errors means that the circle looks funny if you move
  screen shot 2014-11-24 at 12 42 54
- [ ] The readme needs a small section about how to extend the control

src/L.Control.Locate.js
```diff
@@ -130,238 +277,192 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
 
-            resetVariables();
+        /**
+         * Dispatches the 'startfollowing' event on map.
+         */
+        _startFollowing: function() {
+            this._map.fire('startfollowing', this);
+            this._following = true;
+            if (this.options.stopFollowingOnDrag) {
+                this._map.on('dragstart', this._stopFollowing, this, this);
```
domoritz: Why do you pass `this` two times?

gmccrackin: I am also wondering about this.

src/L.Control.Locate.js
```diff
@@ -103,11 +103,158 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
+            return this.options.markerClass(latlng, mStyle)
+        },
+
+        /**
+         * Updates the marker with current coordinates.
+         *
+         * Used by drawMarker, you can ignore it if you have overriden it.
+         */
+        updateMarker: function(latlng, mStyle) {
+            this._marker.setLatLng();
```
gmccrackin: @domoritz @marekkalnik The error looks like it is right here. The setLatLng function is not being passed the latlng variable.

marekkalnik:
Thanks for the feedback! I'll fix the errors this evening.

src/L.Control.Locate.js
marekkalnik: Thanks @gmccrackin. Seems like passing the parameter has indeed fixed it.

gmccrackin: np

src/L.Control.Locate.js
marekkalnik: It was introduced by error. Fixed.

marekkalnik:
I have also added some doc, renamed the methods one would like to call to stop locating and removed some unnecessary parameters (especially in public methods).

I guess it should be ok now.

domoritz:
I was going over it. Will merge your latest changes and then commit to gh-pages. Thanks! This is great.

marekkalnik:
Great. Once it's done I'll try to port jtomaszewski's compass implementation to check if the API is exstensible enough.

domoritz:
Cool! I merged and changed a few minor things.

Issue #102, opened 2014-10-24
keepCurrentZoomLevel not working


1parkplace:
Any idea why? Is there anything conflicting in my options below?

``` js
L.control.locate({
            keepCurrentZoomLevel: true,
            follow: true,
            stopFollowingOnDrag: true,
            metric: false,
            strings: {
                title: "Find my Location",
                outsideMapBoundsMsg: "You seem to be located outside the bounds of this MLS Region."
            }
        }).addTo(this.map);
```

1parkplace:
Not sure if it's related by I get the following error in mapbox.js when I hit my locate buttons:

TypeError: t is undefined
https://api.tiles.mapbox.com/mapbox.js/v2.1.3/mapbox.js
Line 1

domoritz:
It's working for me. Maybe it's a problem with mapbox. Reopen if still an issue.

espy:
I have the same issue (it's a bit of an edge case), and I think I figured it out.

I just saw that `keepCurrentZoomLevel: true` overrides `locateOptions.maxZoom`, but what overrides everything are the [event bounds](https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/src/L.Control.Locate.js#L168), which in my case are enormous, because the location accuracy is very very bad. As in, 30km radius bad.  

**So, in short: if the accuracy is very low, the event bounds are huge, and override any zoom setting.**

This is probably intentional, since you'd want to see the whole accuracy circle overlay after geolocation. Since this also overrides `locateOptions.maxZoom`, there doesn't seem to be a way to stay zoomed in if the GPS accuracy is bad. 

Maybe this helps someone :)

And thanks for the great plugin, @domoritz !

domoritz:
That's an interesting edge case that I haven't thought about.

ghost:
Hi. Do you found a fix for this issue?. Im using mapbox.js and in my site: www.mdqmap.com keepCurrentZoomLevel: true is not working too....

domoritz:
No, I haven't looked at it. I am happy to review pull requests, though.

domoritz:
I think I found it. In https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/src/L.Control.Locate.js#L169 it should be `||`, not `%%`. I have to test it.

domoritz:
@cesarGimenez @espy @1parkplace Should be fixed now. Let me know if you still see this issue.

[...]

PR #117, opened 2015-01-29
Keep zoom level if drawCircle: false and keepCurrentZoomLevel: true

espy:
This checks if `drawCircle: false` and `keepCurrentZoomLevel: true`, and then does a `panTo` on the map instead of `fitBounds`. This covers the very specific edge case of not wanting any accuracy information, while having terrible GPS accuracy.

In short: if `drawCircle: false` and `keepCurrentZoomLevel: true`, we can assume that accuracy info (the circle overlay) isn't wanted or even visible, so we shouldn't be zooming out to accommodate it (regardless of the actual accuracy).

This zooming is totally disorienting on very low accuracy geolocations, which will zoom out massively to fit the huge (in my case, 60km across) _but also invisible_ accuracy circle, going to something as far out as zoom level 9 for no apparent reason. 

See #102 for further info.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -165,11 +165,18 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 if (this._isOutsideMapBounds()) {
                     this.options.onLocationOutsideMapBounds(this);
                 } else {
-                    map.fitBounds(this._event.bounds, {
-                        padding: this.options.circlePadding,
-                        maxZoom: this.options.keepCurrentZoomLevel ?
+                    // If accuracy info isn't desired, keep the current zoom level
+                    if(this.options.keepCurrentZoomLevel && !this.options.drawCircle){
+                        map.panTo([this._event.latitude, this._event.longitude], {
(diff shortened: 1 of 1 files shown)


src/L.Control.Locate.js
```diff
@@ -165,11 +165,18 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 if (this._isOutsideMapBounds()) {
                     this.options.onLocationOutsideMapBounds(this);
                 } else {
-                    map.fitBounds(this._event.bounds, {
-                        padding: this.options.circlePadding,
-                        maxZoom: this.options.keepCurrentZoomLevel ?
+                    // If accuracy info isn't desired, keep the current zoom level
+                    if(this.options.keepCurrentZoomLevel && !this.options.drawCircle){
+                        map.panTo([this._event.latitude, this._event.longitude], {
+                            maxZoom: map.getZoom()
```
domoritz: Can you tell me why you need to provide this? Seems like panning shouldn't affect the zooming at all.

espy: Ah, good point, just a second…

PR #122, opened 2015-05-21
addControl with self reference

antpaw:
I think this is can work as well and has better performance.


antpaw:
I also think this function can be removed
https://github.com/domoritz/leaflet-locatecontrol/blob/c6eb717c0fa2f07395f7014dcaaa191bee8ffaca/src/L.Control.Locate.js#L481
because it does the same

domoritz:
You're right.

Issue #127, opened 2015-06-07
Consider cdn.rawgithub.com


zbycz:
Hi,

consider offering your users link such as
https://cdn.rawgit.com/domoritz/leaflet-locatecontrol/v0.43.0/dist/L.Control.Locate.min.js

instead of 
`<script src="//rawgithub.com/domoritz/leaflet-locatecontrol/gh-pages/dist/L.Control.Locate.min.js" ></script>`. But careful once a branch URL is accessed via cdn.rawgithub, it never expires. See https://rawgit.com/ for details.

Kind regards,
Pavel

domoritz:
Raw github is only for testing and I don't want to maintain bower, npm and rawgit. If you set up the deploy script in a way that it automatically updates the readme, I'd merge. For now this is too many things to support.

PR #128, opened 2015-06-07
Update README.md ISSUE #127

SilentT-FR:
Use CDN Github
Use Master instead of V0.43.0


SilentT-FR:
It's best to use a specific tag or commit hash in the URL (not a branch). Files are cached permanently after the first request. :/ 

so don't use MASTER

Use specific release and udpate on each release

PR #129, opened 2015-06-07
Update README.md ISSUE #127

SilentT-FR:
Patch3

Use v0.43.0 instead of master
recommended by https://rawgit.com/
It's best to use a specific tag or commit hash in the URL (not a branch). Files are cached permanently after the first request.


domoritz:
You shouldn't use rawgit anyway for anything serious and I don't want to keep the readme in sync with the latest version.

PR #132, opened 2015-07-14
send signal when starting and stopping locator

jblarsen:
I would suggest sending a signal when starting and stopping the locator. This provides a clean way to know when the locator is active or not.

(I use it for updating my URL so that if users bookmark my app when the locator is active it will be reflected in the URL)

The same effect can of course be achieved by extending (overriding) the start and stop methods without changing anything in the leaflet-locatecontrol code itself.

Best regards,
Jesper

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -330,6 +330,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
             } else {
                 this.drawMarker(this._map);
             }
+            this._map.fire('startlocator', this);
         },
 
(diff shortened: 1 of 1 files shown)


domoritz:
Yeah, I feel this should be something that you achieve by extending the plugin. I prefer not to bloat the plugin. Sorry.

jblarsen:
No problem. The extension solution is fine with me. Just wanted to let you know if my use case was a common one.

domoritz:
I'll include it if somebody else reports that they would use it.

PR #134, opened 2015-07-20
Update bower.json (dependencies mistake)

cederache:
CF issue "Dependencies mistake"


domoritz:
Thank you. Also solved in #135.

PR #136, opened 2015-08-04
Allow user defined callback "onLocationFound"

exside:
This can be helpful if stuff should be done after the default control stuff is handled...not sure why it's not there, for onLocationError it's there as well, but maybe I miss something here =)


domoritz:
you could subscribe to `map.on('locationfound', fn, this);` but that may be fired before the control is done. Can you add documentation to the readme and also a default implementation in the default options?

domoritz:
Closing because of inactivity.

PR #138, opened 2015-08-20
Add Esri variables

rick-frantz:
Added EsriLeaflet variables


rick-frantz:
This was an accident. Please disregard. If I get it working then I may do an actual pull request but for now, there's nothing to see here.

PR #126, opened 2015-06-06
Advanced configuration. Display control together with zoom.

kolegm:
Added ability to display `locate` control together with `zoom` control.
1. added option to switch the situation. Used `false` by default.
2. check if zoom control displays on a map.
3. corrected css-styles. Leaflet rules.

See attached pic as the working result.
control-locate

PR #142, opened 2015-09-11
displayWithZoomControl

nzmedet:
With `.scss` and `.js` modifications based on this issue https://github.com/domoritz/leaflet-locatecontrol/pull/126, 14 file changed - it's just `grunt` minifications, indeed 2 files changed, also demo for  `demo/index.html` updated. Also removed font-awesome dependency, but this might be redundant, i'm just using another icon plugin


demo/index.html
```diff
@@ -4,9 +4,9 @@
     <meta charset="utf-8" />
     <meta content="width=device-width,initial-scale=1.0" name="viewport" />
 
-    <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css" rel="stylesheet">
+    <link href="http://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css" rel="stylesheet">
```
domoritz, line 7: don't add the protocol to force http

dist/L.Control.Locate.css
```diff
@@ -1,5 +1,5 @@
 /*! Version: 0.44.0
```
domoritz, line 1: As per https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/README.md#developers, please don't include the compiled css.

demo/index.html
nzmedet, line 7: It will not hurt

dist/L.Control.Locate.css
nzmedet, line 1: just add `dist/` to .`gitignore`

demo/index.html
domoritz, line 7: Yes, it does hurt because if somebody uses https, then they cannot request resources via http. If you want, you can force https but don't force http.

dist/L.Control.Locate.css
domoritz, line 1: Good idea. Will do. Nonetheless, can you remove the css from this PR?

domoritz:
Closed due to inactivity.

PR #144, opened 2015-10-31
Add "tagNameForIconElement" option

derzinn:
Make the tag name for the icon element configurable so it can be used with icon sets which don't uses the "span" element.

Working example for [Semantic UI](https://github.com/Semantic-Org/Semantic-UI):

``` js
L.control.locate({
  icon: 'fitted marker icon',
  iconLoading: 'fitted spinner loading icon',
  tagNameForIconElement: 'i'
}).addTo(map);
```


domoritz:
Awesome. Will merge when
- [x] rename to `iconElementTag`
- [x] add documentation to readme

Issue #149, opened 2016-02-11
Option to move marker for location tracking


kentr:
I need a map that follows the user's position on the map without panning the map.  I've had good results with custom code that sets the marker lat / lon in `onLocationFound`.

Would a PR for this feature be welcome?  Perhaps a new option such as `keepCurrentMapBounds` that flags whether `drawMarker()` should `panTo` or instead simply move the marker position?

domoritz:
`setView:false` should already do that. Please let me know if it is not working. 

Or is the difference that you want the map to pan/zoom when the use is first located but not when the location updates? If that's the case, that would definitely be a welcome option. How about you change the `setView` option to take a string `always`, `onLocate`, `never` instead of adding another boolean?

kentr:
Yes, more the latter.  I want the user to be able to pan the map and then keep the current bounds instead of having the map automatically pan back to the current location.  

How does this sound?:
1. `setView` can be `[always | untilPan | false]` (or `[always | untilPan | never]` if you prefer).
2. If `setView == untilPan`:
   - The default behavior of keeping the current location at the map center continues until the user pans the map.
   - Once the user pans the map, the object gets a new second state (something like `_userPanned`)
   - So, if `setView == untilPan && _userPanned == true`, it doesn't `panTo` for new location events.
   - If the user clicks the locate button again, `_userPanned` is reset to `false` and the default behavior renews.

domoritz:
Oh, I used to have an option for something similar. You can find it in the readme `stopFollowingOnDrag`. The difference is that this option also disabled following completely. 

You can get the same behavior with

``` js
var lc = L.control.locate().addTo(map);
map.on('dragstart', lc._stopFollowing, lc);
```

But what you are proposing makes way more sense. The user location keep updating but the map just doesn't pan. In addition to what you proposed, the `followMarkerStyle`/`followCircleStyle` should toggle with `markerStyle`/`circleStyle`. The idea is that the user needs some visual feedback that tells them whether the map will take them out of their current location or not. I also think that `untilPan` should be the default behavior. 

Fantastic! Do you mind sending a pr along? This might actually simplify some code. Don't be afraid to change some existing behavior.

domoritz:
@kentr Are you working on this? I'm happy to review a pr from you for this feature.

kentr:
Yes, I'm hoping to get a PR out soon.

PR #152, opened 2016-03-02
Change setView to respect user panning.

kentr:
PR for #149.

Tested on desktop and mobile (iOS).

I considered adding the `_userPanned` guard to the code that sets `_locateOnNextLocationFound` to `true`, but decided not to because I wasn't clear on the flow / logic.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -77,7 +77,13 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 control.stop();
                 alert(control.options.strings.outsideMapBoundsMsg);
             },
-            setView: true, // automatically sets the map view to the user's location
+            // automatically sets the map view to the user's location.
+            // Possible values: 
+            //  - false: never updates the map view when location changes.
+            //  - always: always updates the map view when location changes.
+            //  - untilPan: (default) like 'always', except stops updating the 
+            //    view if the user has manually panned the map.
+            setView: 'untilPan',
             // keep the current map zoom level when displaying the user's location. (if 'false', use maxZoom)
             keepCurrentZoomLevel: false,
             showPopup: true, // display a popup when the user click on the inner marker
@@ -133,6 +139,8 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 this._map.locate(this.options.locateOptions);
             }
             this._active = true;
+            // Reset _userPanned so that drawMarker() will pan map.
+            this._userPanned = false;
 
             if (this.options.follow) {
                 this._startFollowing(this._map);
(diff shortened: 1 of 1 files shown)


domoritz:
hmm, I tried it and it seems that setting  `setView=false` does not work. 

Moreover, can you also add an option `once`. That makes the map pan to the users location but does not pan the map when the location changes?

domoritz:
Also, the follow option is subsumed by your changes, right? 

Let's delete them then.

src/L.Control.Locate.js
```diff
@@ -396,6 +407,12 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
             if (this.options.stopFollowingOnDrag) {
                 this._map.on('dragstart', this._stopFollowing, this);
             }
+            if (this.options.setView === 'untilPan') {
+                this._map.on('dragstart', function (){
+                  // This listener is on map, not on locateControl.
+                  this.locateControl._userPanned = true;
+                });
```
domoritz, line 428: this breaks. Please write

``` js
                this._map.on('dragstart', function () {
                  this._userPanned = true;
                }, this);
```

kentr:
> this breaks.

Ah, thanks for the correction.  I forgot that I had assigned the control object to `map.locateControl`.

> Also, the follow option is subsumed by your changes, right? 

You may be in a better position to answer that.  I see these pieces, which implies to me that `follow` still has an effect.  My interpretation is that `setView` does nothing if `follow == false`.

``` javascript
            if (this.options.follow && this._following) {
                this._locateOnNextLocationFound = true;
            }
```

``` javascript
            if (this.options.follow) {
                this._startFollowing(this._map);
            }
```

> Moreover, can you also add an option `once`. That makes the map pan to the users location but does not pan the map when the location changes?

So, `setView` can be either `false`, `always`, `once`, or `untilPan` (default)?

domoritz:
> You may be in a better position to answer that. I see these pieces, which implies to me that follow still has an effect. My interpretation is that setView does nothing if follow == false.

I admit that this is a bit confusing right now. I need to refactor this a bit. Following means that we set the view to the users location. I also had to look but it seems that the only way to not change the location of the marker when a new location is found is to set watch to false. 

Have a look at https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/src/L.Control.Locate.js#L368

> So, setView can be either false, always, once, or untilPan (default)?

Yes.

kentr:
@domoritz Please review.

To make the logic more clear, I moved the code which performs the `panTo()` into a handler function `doSetView()`.

Also:
- Added `'once'` option.
- Corrected the bug you found for `_userPanned`.
- Confirmed `setView: false` works as expected.

Regarding `follow`, I think that it's still required b/c it triggers `this._locateOnNextLocationFound` [in `_onLocationFound()`](https://github.com/kentr/leaflet-locatecontrol/blob/gh-pages/src/L.Control.Locate.js#L408).  I'm hesitant to remove it, but I put all of the `setView` guards together in `drawMarker()` so that it would be easier to separate them from `follow` when it's time to remove it.

domoritz:
Thank you! I have to carefully go through the changes.

kentr:
@domoritz Is there anything I can do to help this get merged?

domoritz:
I started working on this but have a deadline at the end of the week. I'll finish it up some time next week.

PR #161, opened 2016-04-27
Make sure this._event is set

ghost:
Urgent! Without this fix, plugin is not functioning when location search has not started. See the default demo: https://www.domoritz.de/leaflet-locatecontrol/demo/ (and try to move the map).

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -262,7 +262,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
         setView: function() {
             if (this._isOutsideMapBounds()) {
                 this.options.onLocationOutsideMapBounds(this);
-            } else {
+            } else if (this._event) {
                 if (this.options.keepCurrentZoomLevel) {
                     this._map.panTo([this._event.latitude, this._event.longitude]);
                 } else {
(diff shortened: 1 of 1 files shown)


kentr:
@juliusosokinas What's the specific problem?  That you can't drag the map until locate is active?

If that is in fact the issue, then I confirmed it exists on Chrome, iOS Cordova / PhoneGap / Ionic.

kentr:
Ah, I see:

```
(index):28 Uncaught TypeError: Cannot read property 'accuracy' of undefined 
http://localhost:8101/bower_components/leaflet.locatecontrol/src/L.Control.Locate.js Line: 284
```

kentr:
See #162 for alternate solution that is smaller and may be easier to review.

ghost:
I do not agree with https://github.com/domoritz/leaflet-locatecontrol/pull/162. This does not solve the problem, which is not `_onDrag`but `_drawMarker` itself. `this._event` can be _undefined_, so trying to retrieve any property from this would end up in error. I see in the code the author checks in some places if `this._event` is being defined, but `_drawMarker` and `setView` are missing checks.

kentr:
After my change, I did not encounter any cases where `this._event` was undefined inside `_drawMarker()`, but maybe @juliusosokinas has identified some. 

## 

Kent Richards

kentr:
Anyway, my intention was only to help get a fix in place more quickly by simplifying the changes so the maintainers would have less to review.

In my testing, #162 solved the problem cases that I was aware of.  

@domoritz I trust you'll do what you think is best.

domoritz:
Oh wow. That's a pretty bad oversight. I'll get a fix in right now.

domoritz:
Okay. I believe the right fix is to bind the drag even when the control is activated and unbind when the control is deactivated. This is definitely cleaner than checking whether the event exists in a bunch of places. However, this still means that you cannot drag as the browser is requesting the location.

I just pushed a hotfix and will push a cleaner fix in a few minutes. Unfortunately I don't think automated tests would have helped here.

ghost:
Thanks.

Please note that this now works:

```
this._event = undefined;

_drawMarker: function() {
    if (this._event.accuracy === undefined) {
        this._event.accuracy = 0;
    }
}

_onDrag: function() {
    if (this._event) {
        // ...
        this._drawMarker();
    }
}
```

But this would break again:

```
this._event = undefined;

_drawMarker: function() {
    if (this._event.accuracy === undefined) {
        this._event.accuracy = 0;
    }
}

// You forget about the issue some time after, or new method added by someone else
_someNewMethod: function() {
    // ...
    this._drawMarker();
}
```

There were already cases with Leaflet itself when similar things happened, and the best solution is to check for `this._event` (or even check if it's an object) before `this._event.accuracy`.

Also, automated tests would catch this, because you'd write tests for `_drawMarker`method only, which would then break if `this._event` is not defined.

domoritz:
Follow up fix was in https://github.com/domoritz/leaflet-locatecontrol/commit/2f73526970ba5c09b1e8d8793cb66c6b4febd585

Let me know if you still see any issues with the code.

PR #165, opened 2016-06-08
The method setStyle no longer exists for a marker.

gmccrackin:
This method no longer exists and it probably didn't need to be there. So I removed it because it will throw an error.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -334,7 +334,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 if (!this._marker) {
                     this._marker = new this.options.markerClass(latlng, mStyle).addTo(this._layer);
                 } else {
-                    this._marker.setLatLng(latlng).setStyle(mStyle);
+                    this._marker.setLatLng(latlng);
                 }
             }
 


domoritz:
Why does it not have to be there? We have to set the updated marker style, right? Why does `setStyle` not return `this` any more? I don't see any issues on http://domoritz.github.io/leaflet-locatecontrol/demo/

domoritz:
The method exists for circle markers, though.

gmccrackin:
I was looking at #169 and I was thinking about the problem. Maybe it should just check to see if its using a marker and have one icon styled for following and one styled for not following. I think that is the easiest way to do styling on the markers now.

What do you think about that?

gmccrackin:
Any thoughts on this?

domoritz:
I think I have a cleaner fix in 6585fdb05db1bcecfb39cabe2c2326062c2eba09. Let me know what you think.

gmccrackin:
Sorry I didnt see this sooner. I think it looks good.

gmccrackin:
I took a second look at this and I think there is a problem with it. Now if you use any type of other markerClass it will never get styled. I think it probably needs to be styled initially and then in the else it needs to check if the style can be updated. Thoughts?
ex.
```javascript
if (this.options.drawMarker) {
    var mStyle = this._isFollowing() ? this.options.followMarkerStyle : this.options.markerStyle;
    if (!this._marker) {
        this._marker = new this.options.markerClass(latlng, mStyle).addTo(this._layer);
    } else {
        this._marker.setLatLng(latlng);
        // If the markerClass can be updated with setStyle, update it.
        if (this._marker.setStyle) {
            this._marker.setStyle(mStyle);
        }
    }
}
```

I' sorry that I didn't see this sooner before you already published 57.

domoritz:
If you have a custom class, you could style it in the constructor. Yes, you would never get the `markerStyle` or `followMarkerStyle` but that seems not like a huge limitation. I agree that your solution is a bit more flexible so I released another version.

gmccrackin:
Thanks!

PR #171, opened 2016-09-22
Add returnToPrevBounds option

ilyankou:


domoritz:
Awesome. What's your use case for this?

src/L.Control.Locate.js
```diff
@@ -196,9 +202,15 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                         break;
                     case 'stop':
                         this.stop();
+                        if (this.options.returnToPrevBounds) {
+                          this._map.fitBounds(this._prevBounds);
```
domoritz, line 207: can you define the variable somewhere (even if just set to null)?

ilyankou:
@domoritz I am working on a map that can display markers and geojson data from Google Sheets:
https://ilyankou.github.io/leaflet-google-sheets/index.html
I thought that Locate would be a nice feature to have, and `returnToPrevBounds` is a quick way to get from the close-up (where-I-am) view back to the "home" screen.

PR #172, opened 2016-09-27
Fix a bug

ilyankou:
I moved `this._prevBounds = null;` to `onAdd` as previously it was in `onClick` and hence lost the value just as the button was clicked.


domoritz:
Thank you for the quick fix!

PR #199, opened 2017-10-21
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that leaflet-locatecontrol gets 689k hits per month on jsDelivr, so I thought you might like this badge.


domoritz (approved):
Oh sweet. Thanks.

PR #212, opened 2018-09-12
untilPanOrZoom

### LINKED ISSUE #210: `following` class not removed when map is zoomed
seems to work fine when i pan, but if i zoom somewhere away from the geolocated location, the `following` class is still present in the DOM.

jayenashar:
Fixes #210

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Possible options are listed in the following table. More details are [in the cod
 |------------|-----------|-------------------|----------|
 | `position` | `string`  | Position of the control | `topleft` |
 | `layer` | [`ILayer`](http://leafletjs.com/reference.html#ilayer)  | The layer that the user's location should be drawn on. | a new layer |
-| `setView` | `boolean`  or `string`  | Set the map view (zoom and pan) to the user's location as it updates. Options are `false`, `'once'`, `'always'`, `'untilPan'`, or [`'untilPanOrZoom'`](#213) | `untilPan` |
+| `setView` | `boolean`  or `string`  | Set the map view (zoom and pan) to the user's location as it updates. Options are `false`, `'once'`, `'always'`, `'untilPan'`, or [`'untilPanOrZoom'`](https://github.com/domoritz/leaflet-locatecontrol/issues/213) | `untilPan` |
 | `flyTo` | `boolean` | Smooth pan and zoom to the location of the marker. Only works in Leaflet 1.0+. | `false` |
 | `keepCurrentZoomLevel` | `boolean`  | Only pan when setting the view. | `false` |
 | `clickBehavior` | `object`  | What to do when the user clicks on the control. Has two options `inView` and `outOfView`. Possible values are `stop` and `setView`. | `{inView: 'stop', outOfView: 'setView'}` |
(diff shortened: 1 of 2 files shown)


src/L.Control.Locate.js
```diff
@@ -52,11 +52,14 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
-             *  - 'untilPan': (default) like 'always', except stops updating the
+             *  - 'untilPan': like 'always', except stops updating the
+             *                view if the user has manually panned the map.
+             *                The map view follows the users location until she pans.
+             *  - 'untilPanOrZoom': (default) like 'always', except stops updating the
              *                view if the user has manually panned the map.
              *                The map view follows the users location until she pans.
              */
-            setView: 'untilPan',
+            setView: 'untilPanOrZoom',
```
domoritz: Let's not change the default.

src/L.Control.Locate.js
```diff
@@ -473,7 +484,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
         },
 
         /**
-         * When the user drags. Need a separate even so we can bind and unbind even listeners.
+         * When the user drags. Need a separate event so we can bind and unbind even listeners.
```
domoritz: Good catch. Can you also fix the second even in this line?

src/L.Control.Locate.js
```diff
@@ -484,6 +495,18 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
             }
         },
 
+        /**
+         * When the user zooms. Need a separate event so we can bind and unbind even listeners.
```
domoritz: typo: even

domoritz (approved):

domoritz:
Thank you @jayenashar

jayenashar:
:(  i don't like your default.  not sure if there's a way to ask the community for their preferred default?

fixed some of your typos, one of which i copied

domoritz:
> i don't like your default

That's okay. I'm happy to change it if there is either a convincing argument or some kind of vote. I'm very careful with changing defaults because it might break people's code. This is why I would like to separate it from this feature PR. Hope this makes sense.

jayenashar:
oh i forgot to update the readme.  i'll do that now.  do you mind if i use a new issue to use for voting?

domoritz:
I was about to say that ;-)

domoritz:
> do you mind if i use a new issue to use for voting

Not at all. That seems like the best place to get feedback about this. 

Rather than a vote I'd like to see comments for or against changing the default. No vote will be representative.

jayenashar:
looks like we both fix the `and` issue.  i'll resolve the conflict

README.md
```diff
@@ -78,7 +78,7 @@ Possible options are listed in the following table. More details are [in the cod
 |------------|-----------|-------------------|----------|
 | `position` | `string`  | Position of the control | `topleft` |
 | `layer` | [`ILayer`](http://leafletjs.com/reference.html#ilayer)  | The layer that the user's location should be drawn on. | a new layer |
-| `setView` | `boolean`  or `string`  | Set the map view (zoom and pan) to the user's location as it updates. Options are `false`, `'once'`, `'always'`, `and 'untilPan'` | `untilPan` |
+| `setView` | `boolean`  or `string`  | Set the map view (zoom and pan) to the user's location as it updates. Options are `false`, `'once'`, `'always'`, `'untilPan'`, or [`'untilPanOrZoom'`](#213) | `untilPan` |
```
domoritz: Hmm, we don't link to PRs for any other property so let's not do it here either.

jayenashar: it's not the PR.  it's the "vote"

jayenashar: the idea is to make the discussion discoverable.  otherwise i imagine people who do want to use it won't know it's available for discussion

Issue #214, opened 2018-11-03
Combined location and direction / compass feature


brendanheywood:
There are other leaflet plugins that add a compass control to the map eg

https://github.com/stefanocudini/leaflet-compass/blob/master/src/leaflet-compass.js

![image](https://user-images.githubusercontent.com/187449/47948440-73213280-df85-11e8-8df0-26beb6ad2c93.png)

But to me it makes more sense UX wise to have the option to combine this with this location control. If this api is available then it's turned at the same time as the gps and updates while ever the gps updates. If it isn't available the gps location appears as it currently does, but where it does appear there would be an extra 3rd marker overlaid over the markerClass with options like compassClass and compassStyle. This would probably need to be inside a proxy marker which handles the rotation automatically.

The default styles I think would be similar to the apple maps, with a little blue arrow

![image](https://user-images.githubusercontent.com/187449/47948462-d0b57f00-df85-11e8-9c88-40ac1e51145f.png)

domoritz:
I like the idea and would be happy to review a pull request.

brendanheywood:
Awesome, would you be open to bundling some minor style changes to the base blue marker style so it works better when the compass arrow is present? The white ring around the marker appears to be the de facto style in all the major map apps and without that contrast the little blue arrow would be hard to see.

domoritz:
Yes, I'm open to that if it looks good and can be styled easily.

brendanheywood:
@domoritz sorry I didn't send this last night: I have a functional wip, but I'm not quite happy with it yet. The main sticking point is the use of a L.Circle for the blue dot instead of a marker which means there is a zoom artifact where it scales with the map and then has to redraw on zoomend. I've currently added the arrow as a path alongside it but I don't like it and think this would be much better if they were both markers, not paths, so they are perfectly synchronized and neither should need a re-draw after a zoom (and so also work nicely with fractional zoom and while you are pinch zooming etc).

So I'm planning to move both of these to new classes which use the divIcon marker, and some glue code so that they accept setStyle and are backward compatible with L.Circle. It's just a little bit more of a change than expected so wanted to make you are broadly ok with that direction. I don't want to pull in any other dependencies.

domoritz:
That sounds great! I'm also okay with breaking backward compatibility in rare cases where it is reasonable. Thank you for taking the lead on this.

domoritz:
Done and released. Thank you!

brendanheywood:
[...]

PR #215, opened 2018-11-05
Add compass heading indicator #214

brendanheywood:


src/L.Control.Locate.js
```diff
@@ -303,6 +316,10 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 this._map.on('locationerror', this._onLocationError, this);
                 this._map.on('dragstart', this._onDrag, this);
                 this._map.on('zoomstart', this._onZoom, this);
+                this._map.on('zoomend', this._onZoomEnd, this);
+                if (this.options.showCompass) {
+                    L.DomEvent.on(window, 'deviceorientation', this._onDeviceOrientation, this);
```
domoritz: Doesn't this need to be `deviceorientationabsolute`?

domoritz: Doesn't this need to be `deviceorientationabsolute`?

src/L.Control.Locate.js
```diff
@@ -104,12 +108,19 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
-                weight: 2,
-                opacity: 0.9,
-                radius: 5
+                fillOpacity: 1,
+                weight: 3,
+                opacity: 1,
+                radius: 7
+            },
+            /** Compass */
+            compassStyle: {
```
domoritz, line 229: Let's make the compass a bit larger or somehow easier to see (maybe white outline?? could be ugly).

screen shot 2018-11-05 at 07 58 30

src/L.Control.Locate.js
```diff
@@ -421,6 +505,43 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
+
+                this._compassHeading = angle;
+                L.Util.requestAnimFrame(this._drawCompass, this);
+            } else {
+                this._compassHeading = null;
+            }
+        },
+
+        /**
+         * If it needs calibration just remove the heading
```
domoritz: "If the compass needs calibration..."

src/L.Control.Locate.js
domoritz: Maybe shadow? Do you have an idea?

domoritz: Maybe we can add some shadow to the circle, make the circle slightly larger and move the direction indicator closer to the blue circle. It might make it easier to see the direction.

src/L.Control.Locate.js
brendanheywood: Hmm, I don't think so. I'm mostly just porting code from that other leaflet plugin which is admittedly a little old, but deviceorientationabsolute doesn't appear to have a proper spec and doesn't seem to have much cross browser support.

https://developer.mozilla.org/en-US/docs/Web/API/Window/ondeviceorientationabsolute

deviceorientation has a proper spec, is 8 years old and has pretty good support. There are implementation differences but they don't seem to matter for our exact use case:

https://caniuse.com/#feat=deviceorientation

domoritz: I'm asking because I read https://developers.google.com/web/updates/2016/03/device-orientation-changes and it says that "Starting with Chrome 50, the degrees included in the DeviceOrientationEvent are by default no longer absolute with respect to the Earth’s coordinate frame." 

I have not yet tested your branch on my phone and I also don't know what the standard is.

brendanheywood: Yeah looks like google half baked that completely. I've been deving this in chrome dev tools and testing on my iphone and feel this is fairly solid there now. I suspect that the absolute events would only be an issue on android and I'm sorry but I don't have a way to test that in the field.

src/L.Control.Locate.js
brendanheywood: The new style for the marker is fairly similar to both the apple maps and google maps apps (on both web and ios at least), and the arrow is based of the apple maps arrow. Google maps has a kinda blue 'tunnel' showing where you are looking which I don't think is as obvious. I don't think this really needs to be much more largest or higher contrast because when you turn around it's clearly moving against the map background and your eye jumps to it easily.

The 2 new marker classes still honor all the same styles as before and someone using this can continue to go a long way customizing them that way before they'd need to either subclass them or replace them. I'm not super happy with the code here but it's not too bad, but there may be some svg styles I've missed (like stroke-dasharray):

https://github.com/domoritz/leaflet-locatecontrol/pull/215/files#diff-351f3423206f4f74f9a9f72e1feae1afR58

src/L.Control.Locate.js
domoritz: I have an android phone and can test it but I suggest you google first what others say about this. I’ll test it before I merge the PR.

src/L.Control.Locate.js
domoritz: The circles seem a bit larger, though so I suggest we adopt that. The syroke width of the accuracy circle is also much thinner and I prefer that as well. I’m happy to make these style changes in a follow up but wanted to flag them here. 

The tunnel also shows the accuracy of the direction, which I think is nice. 

I also noticed that the white part is pulsing on iOS  but I won’t think we need that. I just haven’t noticed it before. 

On android, the circle does not pulse but also seems a bit larger.

src/L.Control.Locate.js
brendanheywood: I've found a few bits of pieces:

https://gist.github.com/mikaelbr/0fed772d49fe655186a4e6ef4b270481#file-orientation-js-L51-L89

https://github.com/ai/compass.js/blob/master/lib/compass.js#L94-L108

But even that is after the chrome 50 release and doesn't seem to handle the absolute stuff. 

I did end up finding a spec which also has some js code for calculating turning the 3 vectors into a proper compass bearing. But I just don't trust any of this as I can't test it myself.

https://w3c.github.io/deviceorientation/#worked-example

Lastly there is a different api which is sorta the same related 

https://developer.mozilla.org/en-US/docs/Web/API/Coordinates/heading

The heading here is not quite what we want and is inferred from large scale moment, so you could be heading north in a car but the device is facing west so this isn't what we want. Most of the time this is null if you are not moving anyway. Even if we did want this we'd need to patch leaflet core as this isn't passed through the normalized locationfound event we get from there.

For now I have left the code as is and not tried to handle anything beyond what I can actually test. Sorry but I've run out of time for this so I can't do too much more. If you are not happy with the android side then I'd suggest just to not show the compass at all if we don't trust the data in certain cases and then build up support as it gets tested.

src/L.Control.Locate.js
brendanheywood: Funnily enough I'd also added the animated pulsing on my locate project which uses this. So I've just added bundled these in as well, I am animating both the marker and the circle. The concept of the 'tunnel' showing the accuracy of the direction I agree is nice, but I think googles styling isn't very good as it is almost invisible on any complex baselayers eg satelite. On top of that I don't think we have any api to return the accuracy of the direction. We could possible make a heuristic for this based on the compass 'jitter' but at least on ios this seems pretty good and I don't think the extra complexity will add a heap of value. This only needs to be vaguely correct within say 10% to be useful.

src/L.Control.Locate.js
brendanheywood: I have made a tiny change, I conditionally listen for the absolute event and only handle either the webkit compass or an absolute event. The chrome dev tools do not actually support this properly, ie it sends either an absolute event with no alpha data, or a relative event which isn't absolute. So at minimum this should ensure we don't show the compass where it is not trusted.

src/L.Control.Locate.scss
```diff
@@ -15,3 +15,24 @@
...
+.leafet-control-locate-location circle {
+    animation: leaflet-control-locate-throb 4s ease infinite;
+}
+@keyframes leaflet-control-locate-throb {
+    0% { radius: 11; stroke-width:1; }
+   50% { radius:  9; stroke-width:3; }
+  100% { radius: 11; stroke-width:1; }
+}
+
+.leaflet-control-locate-circle {
```
domoritz: Why does the accuracy circle pulse?

domoritz: crop

I mean, it's beautiful but also distracting.

src/L.Control.Locate.js
domoritz: I'll test it on my android tonight and merge if it works.

src/L.Control.Locate.js
domoritz: Why are you animating the accuracy circle, though? It seems quite distracting (albeit being beautiful).

src/L.Control.Locate.scss
domoritz: Looks like the NYT is using it on their map right now: https://www.nytimes.com/interactive/2018/11/05/opinion/election-midterms-2018-emotions.html

brendanheywood: In short because it does on ios maps app which I am emulating, and I think it adds value. Happy either way if you don't want it.

I was actually thinking of making it pulse only a single time and leaving it up to the redraw of the circle to kick off the new pulse so that this actually lets you know when the data is refreshed. This didn't work for some reason so I just left it as infinite. If we want to correctly handle a loss of signal then we probably need to handle that in a more robust way in code not css, eg a separate timeout which gets reset with each new location event, and with an alternate set of styles, ie the accuracy marker could be removed or it could turn gray or something. Either way this is beyond the scope of what I wanted to achieve here.

domoritz: Hmm, I don't see the effect on my ipad with apple maps.

brendanheywood: I'm on iphone ios 12. Just confirmed this doesn't show on my ipad either. I've just removed that and squished it back to one commit

domoritz: I watched some videos and they fade it out in the end.

Anyway, for prosperity, here is the CSS

```css
.leaflet-control-locate-circle {
  animation: leaflet-control-circle-pulse 5s ease infinite;
  transform-origin: center;
  transform-box: fill-box;
}

@keyframes leaflet-control-circle-pulse {
    0% { fill-opacity: .4;  transform: scale(.4); }
   20% { fill-opacity: .15; transform: scale(1);  }
   90% { opacity:  1; }
  100% { opacity: .1; }
}
```

domoritz:
Thank you for the PR. I will test it on various devices tonight and merge if it works.

Issue #218, opened 2018-11-07
Add a getZoomBounds callback


brendanheywood:
At the moment the options for setting the zoom level are based purely on the accuracy of the location and not on anything else.

Some use cases:

1) the 'directions' use case: ie I'm in the google maps app and driving somewhere getting turn by turn directions. If you are on the open road you get zoomed out and follow at a high zoom level. But as you approach a feature like a turn or roundabout then the map zooms in, and then zooms out again after the turn.

2) the use case I have is similar, as I am navigating into and out of a nested set of boundaries for polygon features. Depending on where I am in this nested hierarchy I want to show a zoom level which is broadly the same size as that feature.

Suggested solution:

```
getZoomBounds: function(map, latlng, accuracy) {
    ... figure out bounds
    return bounds;
}
```

I can also imagine a use case where that call is async so maybe if you return null then it just updates the pan and it can do whatever it wants later eg:

```
getZoomBounds: function(map, latlng, accuracy) {
    someAjax.then(function() {
        map.fitBounds(bounds);
    });
    return null;
}
```

I think the result of the above would be used exactly and all the other options like maxZoom, circlePadding etc would be ignored as you could build that into your custom function. It would still honor the flyTo vs panTo

domoritz:
`keepCurrentZoomLevel` addresses part of this but I can see the value of more flexibility. I do want to keep the library simple, though.

brendanheywood:
This should almost be a 1 liner here:

https://github.com/domoritz/leaflet-locatecontrol/blob/gh-pages/src/L.Control.Locate.js#L485

If getZoomBounds is set then call is, otherwise use the event bounds. I'll make a pr so you can judge it properly

domoritz:
Wouldn't it be easier for the user to return a zoom level? Can you think of a way to let people do that?

brendanheywood:
[...]

PR #219, opened 2018-11-08
Add getLocationBounds callback #218

brendanheywood:


src/L.Control.Locate.js
```diff
@@ -168,6 +168,19 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
+             * This callback can be used to override the viewport tracking
+             * This function should return a LatLngBounds object.
+             *
+             * For example to extend the viewport to ensure that a particular LatLng is visible:
+             *
+             * getLocationBounds: function(locationEvent, locateControl) {
+             *    return locationEvent.bounds.extend([-33.873085, 151.219273]);
+             * },
+             */
+            getLocationBounds: function (locationEvent, locateControl) {
```
domoritz: isn't `locateControl` just `this` in the function context?

brendanheywood: Yep fixed this up now thanks

domoritz:
Don't you need to now bind this so that one can access the locate control in the callback?

brendanheywood:
No leaflet core handles this for free, `this` is the control

domoritz:
Sweet. I feel at some point I want to move all functions in the options to be just functions of the locate control itself. You could still patch them but they wouldn't be in the options anymore.

Issue #220, opened 2018-11-08
Bug with the first location triggering a follow and then immediate stop-follow


brendanheywood:
Reproduce:

1) map with default options, eg demo map
2) clean page load, press locate button
3) the locate always works but some times the button ends up blue and some times orange

4) If you stop tracking and start again it always works the second time.

Under the hood, it is correctly setting it to blue / following, and then immediately it's reset back to just active. The programmatic call to fitBounds is triggering a zoom event which the locate control thinks is a manual user zoom. The second time the viewport doesn't change so the zoom event isn't triggered.

![image](https://user-images.githubusercontent.com/187449/48193075-bb818b80-e39d-11e8-96df-3fc2bf570c52.png)

I'm happy to fix, just going to set a temp flag before and after and ignore the programmatic zooms.

PR #221, opened 2018-11-08
Fixes follow reset bug #220

brendanheywood:


src/L.Control.Locate.js
```diff
@@ -482,10 +482,12 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                     f.bind(this._map)([this._event.latitude, this._event.longitude]);
                 } else {
                     var f = this.options.flyTo ? this._map.flyToBounds : this._map.fitBounds;
+                    this._ignoreEvent = true;
```
domoritz, line 499: Please add a comment about why this is necessary.

domoritz (changes requested):

brendanheywood:
Hmm this fix isn't good enough, depending on the timing of the requestAnimationFrame the subsequent zoom and zoom event could be async in the next tick which explains why this was intermittent.

brendanheywood:
Ok this should be pretty solid now. In my research I also found another concept:

https://stackoverflow.com/questions/31992278/detect-user-initiated-pan-zoom-operations-on-leaflet

But that e.hard attribute was never documented and has since been removed from leaflet core

domoritz:
Thank you!

Issue #225, opened 2018-11-12
Zooming via mouse wheel can effectively be a pan but doesn't trigger _userPanned


brendanheywood:
If you zoom using the mouse wheel then it centers the zoom where the cursor is. This can end up with some funny side effects, when setView == 'untilPan'.

I can zoom out of one location, and then zoom in somewhere else. This is exactly equivalent to a pan, except a pan event never gets fired. I am still in following mode, even though the location marker may not even be inside the view port anymore.

This is admittedly a very small edge case, especially considering people with mouse wheels are typically on desktop and probably don't have gps so maybe it's not worth handling.

But if we handle this then I'd suggest there is some sort of tolerance and if in the act of zooming the center moves than some % of the viewport width then it is treated as a pan and following is turned off. Or perhaps the simpler option of if after a zoom the marker is outside the viewport then stop following.

domoritz:
Yeah, I noticed this before but it's not easy to solve. Maybe it makes sense to change the default so that zooming also disabled following. What do you think?

brendanheywood:
Well the default is setView: untilPanOrZoom so that's fine and I don't think we should change anything there. I think I'll just let this one sit for a while until an obvious solution shows itself

domoritz:
Right. Well, maybe we should just delete `untilPan` then as it doesn't really work anyway???

brendanheywood:
I'm confused, in my mind untilPan works perfectly and it's what I've set my app to use. A pan via a drag turns off following as expected, but a zoom doesn't, which is expected IF it is centered in the viewport. With this patch now a sideways zoom via a mousewheel also triggers a pan and turns off following.

There are other ways you could end up panning without this control detecting it. eg its possibly could also apply if leaflet core or another plugin added a touch gesture way to zoom out, eg a single finger tap zoom in, two finger tap zooms out (google and apple maps do this)

I think I've found a semi related bug, which is if the map is at max zoom, and you double touch tap, then the map cannot zoom in, but it will re-center on where the tap does which is effectively a pan - but the following doesn't turn off when it should. Leaflet is firing a moveend but not a zoom* or drag* event.

Anyway you've merged that pr so I'll close this one

btw I am getting a little anxious about changes without tests :/

domoritz:
Ignore me. I merged your PR. 

> btw I am getting a little anxious about changes without tests :/

Me too... I just don't have the time right now.

PR #226, opened 2018-11-13
Detect zooms that are effectively pans #225

brendanheywood:
.3 is chosen so it's roughly 50% of the viewport width


domoritz:
This is great! Thanks.

Issue #234, opened 2019-02-12
this._marker is undefined when drawMarker is false and map zoom


SuperPat45:
I have the exception when the drawMarker is at false and the map zoom:
> TypeError: this._marker is undefined

At the line:
`if (!this._map.getBounds().pad(-.3).contains(this._marker.getLatLng())) `

PR #235, opened 2019-02-12
this._marker is undefined when drawMarker is false and map zoom

SuperPat45:
There are an exception when the drawMarker is at false and the map zoom:

> TypeError: this._marker is undefined

At the line:
`if (!this._map.getBounds().pad(-.3).contains(this._marker.getLatLng())) `

Fix bug #234


domoritz:
Thanks!

PR #247, opened 2019-07-16
Add zoom range restriction to keepCurrentZoomLevel

### LINKED ISSUE #246: Enhancement Request: keepCurrentZoomLevel within set range
`keepCurrentZoomLevel` in general is a nice option. 

But think it would be even better to have a middle ground. Eg only keepCurrentZoomLevel if the current zoom is between say 13 and 18. 
... so if below 13 (zoomed out) would zoom in. 
... and if zoomed in beyond 18 it could zoom out (assuming maxZoom was set to say 16) 

This would give best of both worlds? If the user is zoomed way out (say country level) then it would zoom in to show the proper position. But if they've already set to a nice midscale, it would keep it, rather than switching to the exact option the developer choose.

barryhunter:
Fixes https://github.com/domoritz/leaflet-locatecontrol/issues/246


README.md
```diff
@@ -80,7 +80,7 @@ Possible options are listed in the following table. More details are [in the cod
 | `layer` | [`ILayer`](http://leafletjs.com/reference.html#ilayer)  | The layer that the user's location should be drawn on. | a new layer |
 | `setView` | `boolean`  or `string`  | Set the map view (zoom and pan) to the user's location as it updates. Options are `false`, `'once'`, `'always'`, `'untilPan'`, or `'untilPanOrZoom'` | `'untilPanOrZoom'` |
 | `flyTo` | `boolean` | Smooth pan and zoom to the location of the marker. Only works in Leaflet 1.0+. | `false` |
-| `keepCurrentZoomLevel` | `boolean`  | Only pan when setting the view. | `false` |
+| `keepCurrentZoomLevel` | `boolean` or `Array` | Only pan when setting the view. Set to True, or a zoom range like [13,18] which will only zoom the map when outside the rage. Allows keeping the current view at these scales | `false` |
```
domoritz: Please correct the grammar of this sentence. Also, end with a `.`.

src/L.Control.Locate.js
```diff
@@ -514,7 +514,10 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 this._event = undefined;  // clear the current location so we can get back into the bounds
                 this.options.onLocationOutsideMapBounds(this);
             } else {
-                if (this.options.keepCurrentZoomLevel) {
+                if (this.options.keepCurrentZoomLevel && (
+                    typeof this.options.keepCurrentZoomLevel == "boolean" ||
+                    (typeof this.options.keepCurrentZoomLevel == "object" && this._map.getZoom() <= this.options.keepCurrentZoomLevel[1] && this._map.getZoom() >= this.options.keepCurrentZoomLevel[0])
```
domoritz, line 519: Can you extract this into a variable `zoomOutsideRange` and then write `if (this.options.keepCurrentZoomLevel || zoomOutsideRange)`?

domoritz (changes requested):
Thank you for the PR. I have two small comments but would be happy to merge once you addressed them.

README.md
barryhunter: Not sure what the grammer issue is, but have just spotted the rage=>range typo. I'll try fixing this - I think as a new PR.

src/L.Control.Locate.js
barryhunter: Not sure what mean, make something like 

      var zoomOutsideRange = typeof this.options.keepCurrentZoomLevel == "object" && this._map.getZoom() <= this.options.keepCurrentZoomLevel[1] && this._map.getZoom() >= this.options.keepCurrentZoomLevel[0];

if (this.options.keepCurrentZoomLevel)  would still be true in case of Array, so the if(..) still complicated. 

although maybe could do `if (this.options.keepCurrentZoomLevel === true || zoomOutsideRange)`? (not used === in javascript!) 

---

Or to have new `this.options.zoomOutsideRange` to store the range? 

      var zoomOutsideRange = typeof this.options.zoomOutsideRange == "object" && this._map.getZoom() <= this.options.zoomOutsideRange[1] && this._map.getZoom() >= this.options.zoomOutsideRange[0];

Then could just do... keepCurrentZoomLevel is back to just true/false

    if (this.options.keepCurrentZoomLevel || zoomOutsideRange )

README.md
domoritz: Something like this (it wasn't clear what the which clause refers to):

```suggestion
| `keepCurrentZoomLevel` | `boolean` or `Array` | Only pan when setting the view. Setting a zoom range such as `[13,18]` will only zoom the map if the current zoom level is outside the specified range. | `false` |
```

domoritz:
Hmm, thinking about this a bit more, I am not sure we want to merge this yet. When I see an array for `keepCurrentZoomLevel`, I'd think that this is providing the range that the zoom should be in (i.e. the target zoom level). I feel that filtering when to zoom is something that is very specific and should be done outside. 

I'd say let's add an event that gets triggered so that you can implement your own logic for zooming in, okay? 

I'm going to close this PR for now. 

I hope you understand that I want to keep only core functionality as otherwise this library will get too complex to maintain.

README.md
barryhunter: just saw reply. I've updated the wording, try to keep clearer. It only zooms when outside the range. It keeps the zoom when within the range! ie keep zoom (if True, or within set range) https://github.com/barryhunter/leaflet-locatecontrol/commit/1eaee571533969af8c888b7b7d745490e4bfb255 
not clear if that commit, is feeding back to pull request. if not will make a new one, after resolving other comment)

barryhunter:
ok. sure

but *I'd think that this is providing the range that the zoom should be in (i.e. the target zoom level).*

that is what is doing. 

If define as `[13,18]` then, if outside that range, it WILL zoom to that range. Its only if its already within 13->18 that it doesnt zoom. ie keep current zoom, if already within 13->18.

domoritz:
Not quite. What I thought a range would be a is a range to clamp to. For example, if the user defines `[13,18]`, then the zoom level would be between 13 and 18 but we might still zoom. But you are right that this isn't quite compatible with `keepCurrentZoomLevel` as a name, which implies that the range has something to do with keeping the zoom level. I think now I understand it better and updated the wording a bit. Now I also see that this is quite useful and so let's merge it. 

Sorry, I needed a moment to understand this proposal fully.

domoritz:
Could you do these two things to finish the PR?

- [ ] Have a look at https://github.com/domoritz/leaflet-locatecontrol/pull/247#discussion_r303865857 and see whether you want to apply the suggestion. 
- [ ] Apply your refactoring to extract `zoomOutsideRange` as you discussed in zoomOutsideRange. 

Thanks!

barryhunter:
So are thinking to have as a new separate 'option', as opposed to overloading keepCurrentZoomLevel. 
ie add 'options.zoomOutsideRange'. 

Or is it just reflectoring the complicated if(..) clause that concerned about?

domoritz:
I think keeping it part of `keepCurrentZoomLevel` is fine (no separate option). I am concerned about the docs (they need to be crystal clear so that others don't run into my confusions ;-)) and the complicated `if` clause.

domoritz:
Done in main now.

PR #255, opened 2019-10-24
iOS 13: deviceorientation error for showCompass

sveine:


src/L.Control.Locate.js
```diff
@@ -467,10 +467,27 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
-                        L.DomEvent.on(window, 'deviceorientation', this._onDeviceOrientation, this);
+                    var oriAbs = 'ondeviceorientationabsolute' in window;
+                    var oriStd = 'ondeviceorientation' in window;
+                    if (oriAbs || oriStd) {
+                        var _this = this;
+                        var deviceorientation = function () {
+                            if (oriAbs) {
+                                L.DomEvent.on(window, 'deviceorientationabsolute', _this._onDeviceOrientation, _this);
+                            } else if (oriStd) {
+                                L.DomEvent.on(window, 'deviceorientation', _this._onDeviceOrientation, _this);
```
domoritz: Can you only set the second argument based on the oriAbs or oriStd?

src/L.Control.Locate.js
```diff
@@ -467,10 +467,27 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
+                            }
+                        };
+                        if (DeviceOrientationEvent && typeof DeviceOrientationEvent.requestPermission === 'function') {
+                            DeviceOrientationEvent.requestPermission().then(function (permissionState) {
+                                deviceorientation();
+                                if (permissionState === 'granted') {
+                                    deviceorientation();
```
domoritz, line 479: Why do we call the same function twice if granted is true?

domoritz (commented):
Thank you. I have two comments.

src/L.Control.Locate.js
sveine: Agree!

src/L.Control.Locate.js
sveine: The statement was for debugging: removed

PR #258, opened 2019-12-26
Support functions for `strings.popup`

simon04:
This allows to use an external i18n system with its argument and plural handling.

As a not very serious example for testing, you may want to try:

```javascript
{
  strings: {
    popup: args => `You are witih half of ${args.distance * 2} ${args.unit}`
  }
}
```


domoritz:
Can you update the readme as well?

Am I missing something or does this only support functions for one of the strings?

simon04:
README: done in https://github.com/domoritz/leaflet-locatecontrol/pull/258/commits/c3662e5fd2ef37368bfb2ea2edcdeec96eab6c36

Only `strings.popup`: Yes, since it is the only string that deals with arguments/placeholders. I don't see a point to evaluate the other strings via a function.

PR #263, opened 2020-03-21
Option 'initialZoomLevel': zoom to specified zoom level after click

plepe:
This pull requests adds an option 'initialZoomLevel', which will zoom to the specified zoom level to zoom to after clicking on the icon (even when keepCurrentZoomLevel is `true`).

In my app I want to zoom to a specified zoom level when acquiring a position, but when the user changes the zoom while the plugin is active, it should not change zoom any more.


domoritz:
Thank you for the pull request. Can you update the readme as well?

plepe:
You are welcome :-) I just updated the README and force-pushed the commit. Any comments / thoughts?

plepe:
The change is already used by https://www.openstreetbrowser.org/ :-)

domoritz:
Awesome, thank you for the code and using the plugin.

PR #264, opened 2020-04-02
Set this._map in onAdd for controls outside the map

FlorianGerhardt:
I'm trying to use this plugin outside of the map so I don't call addTo(map).

When I use onAdd() this._map is not set, so this PR sets it.

Example from leaflet-control-geocoder: https://github.com/perliedman/leaflet-control-geocoder/blob/b8ebba11dd04a06576c1a67f0d302bc9e7ec7b44/src/control.js#L39

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -335,6 +335,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
         onAdd: function (map) {
             var container = L.DomUtil.create('div',
                 'leaflet-control-locate leaflet-bar leaflet-control');
+            this._container = container;
             this._map = map;
             this._layer = this.options.layer || new L.LayerGroup();
             this._layer.addTo(map);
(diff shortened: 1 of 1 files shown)


FlorianGerhardt:
I also added setting this._container in onAdd so classes get set correctly.

domoritz:
Thanks for the pull request. Can you explain the fix to me a bit more? 

You say that you don't call `addTo(map)` but the function you are editing calls it for you, no? 

Screen Shot 2020-04-02 at 10 14 10

FlorianGerhardt:
So I'm trying to use this plugin outside of the map. I managed to implement this by following this SO [post](https://stackoverflow.com/questions/59743050/react-leaflet-placing-map-control-components-outside-of-map)

So to summarize the PR:
I dont call .addTo(map) on the locatecontrol itself so it doesnt show up inside the map but instead append the created div to a predefined container.

const locatecontrol = new Locatecontrol();
const locateContainer = locatecontrol.onAdd(map);
container.current.appendChild(locateContainer);

Thats why this._map and this._container are not set as the plugin is not setup the intended way.

Edit: L341 you are marking only adds the layer which shows the marker / radius on the map, afaik.

domoritz:
Sounds good. Thanks for the explanation. 

Alles Gute nach Berlin!

FlorianGerhardt:
Bleib gesund! :)

Issue #268, opened 2020-05-08
Accessibility - Keyboard navigation


Michael-Hanley:
Hello, 

Thanks for creating this awesome project! it's been incredibly useful. 

The issue I have been facing is related to the accessibility of the locate control. It is not included in the tab order so it cannot be focused on or acted on with a keyboard.

domoritz:
Thanks for noting the issue. Can you send a pull request to fix it?

Michael-Hanley:
Hello, 

I've opened a PR with a fix. 

Thanks

domoritz:
Will be fixed in https://github.com/domoritz/leaflet-locatecontrol/pull/269

PR #269, opened 2020-05-08
Feature/accessibility

Michael-Hanley:
- Adds locate control to tab index
- Adds keyup events
- Renames _onClick to _onButtonAction and updates comment

Issue: #268

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -280,7 +280,6 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
             createButtonCallback: function (container, options) {
                 var link = L.DomUtil.create('a', 'leaflet-bar-part leaflet-bar-part-single', container);
                 link.title = options.strings.title;
-                link.tabIndex=0;
                 var icon = L.DomUtil.create(options.iconElementTag, options.icon, link);
                 return { link: link, icon: icon };
             },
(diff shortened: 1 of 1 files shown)


src/L.Control.Locate.js
```diff
@@ -361,49 +365,51 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
-                            var f = this.options.flyTo ? this._map.flyToBounds : this._map.fitBounds;
-                            f.bind(this._map)(this._prevBounds);
-                        }
-                        break;
-                }
-            } else {
-                if (this.options.returnToPrevBounds) {
-                  this._prevBounds = this._map.getBounds();
+        _onButtonAction: function(e) {
+            if (e.type === 'click' || e.key === "Enter" ){
```
domoritz: I have not seen this pattern before. Can you point to some code in Leaflet that does the same for the standard buttons?

src/L.Control.Locate.js
```diff
@@ -280,6 +280,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
             createButtonCallback: function (container, options) {
                 var link = L.DomUtil.create('a', 'leaflet-bar-part leaflet-bar-part-single', container);
                 link.title = options.strings.title;
+                link.tabIndex=0;
```
domoritz: So if we are 0, how do we make sure that other buttons come before in the tab order?

domoritz (changes requested):

domoritz:
I don't see a tab index for the other buttons. Why do they work already?

Screen Shot 2020-05-08 at 17 04 54

src/L.Control.Locate.js
Michael-Hanley: tabindex="0" means that the element should be focusable in sequential keyboard navigation, after any positive tabindex values and its order is defined by the document's source order.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

But it should be a char not a number, I'll make that change.

Michael-Hanley:
Thanks for pointing all that out, turns out this fix was much simpler.

src/L.Control.Locate.js
Michael-Hanley: Removed

src/L.Control.Locate.js
Michael-Hanley: Removed

domoritz:
Awesome. Thanks for making the change so small.

Issue #281, opened 2021-03-12
Start and stop events


tgelz:
I recently added a feature to our Leaflet application that required knowing when the locate control was being activated or deactivated to trigger certain changes in the map's behavior. There currently doesn't seem to be a proper way to achieve this (unless I'm missing something), so I extended the control to be able to fire events.

It would be nice if these events could be added as a standard feature, as extending a control and overriding native functions means you always have to check if something changed after an update.

I placed `this._map.fire('locatestart')` and `this._map.fire('locatestop')` after setting the `_active` flag in the `_activate()` and `_deactivate()` functions respectively, but there might be a better place for these.

domoritz:
Can you use the createButtonCallback and add a click listeners to the link? Or is that not enough?

tgelz:
Unfortunately, that wouldn't be reliable enough as the locate control can (and does) get started and stopped programmatically, not only by our code (e.g. autostart on page load), but also by things like `onLocationOutsideMapBounds` and ` _onLocationError`.

domoritz:
Makes sense. Can you send a pull request?

PR #282, opened 2021-03-12
Add activate and deactivate events

tgelz:
#281

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -468,6 +468,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
         _activate: function() {
             if (!this._active) {
                 this._map.locate(this.options.locateOptions);
+                this._map.fire('locatestart', this);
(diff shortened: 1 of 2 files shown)


domoritz (changes requested):
please update the docs

src/L.Control.Locate.js
```diff
@@ -468,6 +468,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
         _activate: function() {
             if (!this._active) {
                 this._map.locate(this.options.locateOptions);
+                this._map.fire('locatestart', this);
```
domoritz: Since this function is called `activate` do you think it would be better to call the even `locateactivate`? Similarly for the other one.

tgelz: Hmm, I was originally basing this on the `start` and `stop` functions for analogy (which in turn call `activate` and `deactivate`), but I think you're right and it would be more accurate and representative to rename the events.

PR #287, opened 2021-06-24
Support customization of link text

yanis-sadeg:
## Description
Added functionality to add text to the location button. 
It is activated with the "text" option and does not allow HTML characters.

Tested on Firefox, Chrome, Safari. ✔️

## Example
```
lc = L.control.locate({
    strings: {
        title: "Show me where I am, yo!",
        text: "Locate me"
    }
}).addTo(map);
```

## Screenshot
Leaflet locate link text


README.md
```diff
@@ -213,7 +213,7 @@ var map = new L.Map('map', {
 
 ## Developers
 
-Run the demo locally with `grunt serve` and then open [localhost:9000/demo/index.html](http://localhost:9000/demo/index.html).
+Run the demo locally with `grunt connect` and then open [localhost:9000/demo/index.html](http://localhost:9000/demo/index.html).
```
domoritz:
```suggestion
Run the demo locally with `yarn start` or `npm run start` and then open [localhost:9000/demo/index.html](http://localhost:9000/demo/index.html).
```

src/L.Control.Locate.js
```diff
@@ -283,6 +285,14 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 link.role = 'button';
                 link.href = '#';
                 var icon = L.DomUtil.create(options.iconElementTag, options.icon, link);
+
+                if(options.strings.text !== undefined) {
```
```suggestion
                if (options.strings.text !== undefined) {
```

src/L.Control.Locate.js
```diff
@@ -283,6 +285,14 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
+                    var text = L.DomUtil.create(options.textElementTag, 'leaflet-locate-text', link);
+                    text.textContent = '\xa0' + options.strings.text;
```
domoritz: Why do you need to add `\xa0`?

src/L.Control.Locate.js
```diff
@@ -283,6 +285,14 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
...
+		    link.classList.add('leaflet-locate-text-active');
```
domoritz: indentation?

src/L.Control.Locate.scss
```diff
@@ -15,6 +15,15 @@
   }
 }
 
+.leaflet-touch .leaflet-bar .leaflet-locate-text-active  {
```
domoritz: What does this have to do with the proposed change?

domoritz (approved):
Thanks for the pull request. Can you add a screenshot to the description?

src/L.Control.Locate.js
yanis-sadeg: > 
> 
> Why do you need to add `\xa0`?

To make a space between the icon and the text, otherwise they are stuck together and it's ugly. I had the choice to do it by css or like this, i choosen this method. But thinking about, maybe I should make the space by css on the icon with a padding right, like this if the developer choose to use only the text without the icon there is no space for nothing. What do you think?

src/L.Control.Locate.js
yanis-sadeg: > 
> 
> indentation?

Yes... In my side it's correctly indented, I don't know why the indentation on this preview is wrong. I will check about it for the next push.

src/L.Control.Locate.scss
yanis-sadeg: This allows to modify the css only if the text option is used, it will not impact the button if the option is not used. Also, I don't use the .leaflet-control-locate because otherwise I would have to use !important on the width property.

Modifications have been made on the css if the text option is displayed because of the default properties of leaflet which made it impossible to display the text correctly.

yanis-sadeg:
> 
> 
> Thanks for the pull request. Can you add a screenshot to the description?

Sure, I will on the next push.

Edite: The requested changes have been made.

domoritz:
I meant a screenshot in the pull request description not the code. I just want some documentation for the feature here. Sorry for being ambiguous.

yanis-sadeg:
> 
> 
> I meant a screenshot in the pull request description not the code. I just want some documentation for the feature here. Sorry for being ambiguous.

My bad, I had misunderstood. Done 👌

src/L.Control.Locate.js
domoritz: Yep, let's do it in CSS.

src/L.Control.Locate.js
domoritz: Make sure you use spaces.

src/L.Control.Locate.js
```diff
@@ -283,6 +285,17 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
                 link.role = 'button';
                 link.href = '#';
                 var icon = L.DomUtil.create(options.iconElementTag, options.icon, link);
+
+                if (options.strings.text !== undefined) {
+                    var text = L.DomUtil.create(options.textElementTag, 'leaflet-locate-text', link);
+                    text.textContent = options.strings.text;
+		            link.classList.add('leaflet-locate-text-active');
+                    link.parentNode.style.display = "flex";
+                    if(options.icon.length > 0) {
```
```suggestion
                    if (options.icon.length > 0) {
```

domoritz:
Thank you for the addition!

PR #296, opened 2021-11-26
Replace External Font Awesome with Local SVG

### LINKED ISSUE #293: FontAwesome 5
Is there any way of getting this to work with FontAwesome 5? I've tried a few things with limited success.
I'm specifically using FontAwesome 5's VueJs component.

What I thought would work is just enabling the processing of `i` tags, as mentioned here: https://github.com/FortAwesome/vue-fontawesome#processing-i-tags-into-svg-using-font-awesome
And then changing the icon class names to the FontAwesome 5 equivalents.
This partially works, and outputs the correct icon in the initial sate.

However, the icon doesn't get updated with the loading spinner, when it is waiting for permission and finding the location. It also doesn't "highlight" the icon once the position is found.
I think this is because FontAwesome replaces the icon DOM element, with a new `svg` element, and so the class changes don't have an element to modify any more (Or at least not one attached to the DOM).
What I think it needs to do, is to re-build the icon element when the state changes, so it picks up the changes, I think.

Any help would be appreciated.

hswong3i:
Existing implementation reuse Font Awesome icons and spinner animation effect for simpler implementation. But this also lead to additional dependency which make 3rd party integration become harder.

For example, in case of integrating Leaflet.Locate into [Drupal Leaflet](https://www.drupal.org/project/leaflet) module, the use of Font Awesome for icon introduce additional dependency to the needs of [Drupal Font Awesome](https://www.drupal.org/project/fontawesome).

Sometime user may even hope to use a newer/older version of Font Awesome in their project, different than that well tested by ourselves.

This PR decouple above issue by:

  - Locally vendor Font Awesome v5.15.4 `location-arrow-solid.svg` and `spinner-solid.svg`
  - Implement our own independent SCSS as similar as original design provided by Font Awesome
  - Simplify demo and README.md with above changes

Online demo: https://drustack.github.io/domoritz-leaflet-locatecontrol/demo/

Fix #293

Signed-off-by: Wong Hoi Sing Edison <hswong3i@pantarei-design.com>


hswong3i:
Online demo: https://drustack.github.io/domoritz-leaflet-locatecontrol/demo/

src/L.Control.Locate.scss
```diff
@@ -1,18 +1,45 @@
...
-    a {
-      color: #2074b6;
+
+    .leaflet-control-locate-location-arrow {
+      display: inline-block;
+      width: 16px;
+      height: 16px;
+      margin: 7px;
+      background-color: black;
+      -webkit-mask-image: url('../location-arrow-solid.svg');
```
domoritz, line 12: I don't think we need these vendor specific commands anymore. Let's remove them.

README.md
```diff
@@ -236,3 +232,5 @@ To all [contributors](https://github.com/domoritz/leaflet-locatecontrol/contribu
 ## License
 
 MIT
+
+SVG icons from [Font Awesome v5.15.4](https://github.com/FortAwesome/Font-Awesome/releases/tag/5.15.4): [Creative Commons Attribution 4.0](https://fontawesome.com/license)
```
domoritz:
```suggestion
SVG icons from [Font Awesome v5.15.4](https://github.com/FortAwesome/Font-Awesome/releases/tag/5.15.4): [Creative Commons Attribution 4.0](https://fontawesome.com/license/free).
```

domoritz (approved):
Thank you. Just a few small changes.

src/L.Control.Locate.scss
hswong3i: We do need it: Firefox working very well without `-webkit-` prefix, but chrome not, see https://www.w3schools.com/cssref/css3_pr_mask-image.asp

README.md
hswong3i: Done

src/L.Control.Locate.scss
domoritz: Thanks for checking. Let's leave it then.

PR #311, opened 2022-08-15
Corrected README event names

FSeixasPT:
Corrected README event names for the Leaflet [native events](https://leafletjs.com/reference.html#map-locationerror).


domoritz:
Thank you

PR #316, opened 2022-10-23
Create dependabot.yaml

KristjanESPERANTO:
Since npm is only used for development, monthly checks should be sufficient.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


domoritz:
Thanks for the pull request. This library is more or less complete and so I don't think I need to update dev dependencies. If you feel strongly about this, could you add a GitHub action file to at least run the builds? Otherwise I would be worried to introduce regressions.

KristjanESPERANTO:
Thank you for the fast reaction. I think it's desirable to use the latest versions of the dependencies in the build process. And I thought this approach made sense, but if you see it differently, I'm not disappointed :slightly_smiling_face: 

Do you see a better approach to ensure every build process is using the latest dependency versions? Maybe put `npm-check-updates` at the beginning of the build process?

domoritz:
I honestly don't see the need to update the Dev dependencies. They work so I would rather just keep them as is and update manually.

KristjanESPERANTO:
For me it's okay that it stays the way it is :slightly_smiling_face:

PR #317, opened 2022-10-23
Leaflet 1.9.2 and minor changes

KristjanESPERANTO:
- Tested with Leaflet 1.9.2
- Demo with Leaflet 1.9.2
- Screenshot with default icon
- var -> const, let


domoritz:
Thank you for the pull request. Please remove the code style changes and submit them as a separate pull request.

KristjanESPERANTO:
OK, I'll do that.

Do you prefer single or double quotes? If you don't have a preference, I would use double quotes as the default.

KristjanESPERANTO:
And do you have an indentation preference? 2 or 4 spaces?

KristjanESPERANTO:
I just undid the code style changes.

domoritz:
Let's use prettier and its defaults.

KristjanESPERANTO:
I like prettier defaults :slightly_smiling_face: 

So #318 is my PR regarding the code style.

PR #326, opened 2022-11-01
Add dark mode support

### LINKED ISSUE #308: Icon in dark mode
On a phone in Chrome forks with a dark mode (e.g. Adblock browser), the icon becomes invisible after changing the mode - the icon remains black, while other OSM / Leaflet icons can change the colors to light. Is it because of the use of JS and is there any solution?

KristjanESPERANTO:
This fixes issue #308.

Test setup:
1. Open `chrome://flags/` in Chrome
2. Enable `Auto Dark Mode for Web Contents`

| before | after |
| --- | --- |
| before | after |

Spinner, Following and Active also work:

spinner following active 

Also works with `Mapbox.js` :smiley:


domoritz:
Awesome. I'll make a release.

PR #327, opened 2022-11-02
Demo improvements

KristjanESPERANTO:
Detail improvements on the demo pages. The CSS ribbon with hover effect is probably the main thing that catches the eye.


domoritz:
Thank you

PR #330, opened 2022-11-30
fix stopLocate on null issue

### LINKED ISSUE #329: calling stopLocate on null
It is possible to call Stop() on the control after the map is destroyed which causes an error.

See https://github.com/runette/ngx-leaflet-locate/issues/4 for some discussion.

This could be prevented by check for the existance of the map before calling stopLocate.

runette:
fixes #329


runette:
fixes https://github.com/runette/ngx-leaflet-locate/issues/4

domoritz:
Thanks. Fixed in v0.79.0

PR #343, opened 2023-10-03
added oneshot callback (trigger, then stop) when location is found

valerino:


src/L.Control.Locate.js
```diff
@@ -768,9 +777,12 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
       }
 
       this._event = e;
-
       this._drawMarker();
       this._updateContainerStyle();
+      if (this._loc_callback && this._active) {
+        // call the location callback
+        this._loc_callback([e.latlng.lat, e.latlng.lng])
```
domoritz: I think we should pass e.latlan instead.

src/L.Control.Locate.js
```diff
@@ -756,7 +765,7 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
     /**
      * Stores the received event and updates the marker.
      */
-    _onLocationFound(e) {
+    async _onLocationFound(e) {
```
domoritz: Why make this async?

src/L.Control.Locate.js
```diff
@@ -347,6 +350,12 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
         }
       }
 
+      this._loc_callback=null
+      if (options.callback) {
+        // set location callback
```
```suggestion
```

src/L.Control.Locate.js
```diff
@@ -153,6 +153,9 @@ You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
 
   const LocateControl = L.Control.extend({
     options: {
+      /** function([lat,lng]) to be called when location is found.
+      if this is set, stop() is called when location is first found. */
```
domoritz, line 157: Why would be call `stop()` when this callback is set? Maybe explain what the purpose of this API is.

domoritz (changes requested):

src/L.Control.Locate.js
valerino: sorry, typo!

src/L.Control.Locate.js
valerino: because i wanted it to be oneshot (just find my location and quit). either, the handler remains active and the callback would be called again and again until, indeed, stop() is called.  i documented this in the readme, and renamed callback to "oneshot_callback"

domoritz: So why can this not be done from outside the plugin? You can listen to the location found event on the map I think.

domoritz:
Thanks for the updates. Why is this functionality required to be inside the plugin rather than something controlled from the outside (by listening to location updates on the map and then calling stop on the plugin)?

urb4nc4rl:
I'm not the OP but I also came here looking for this functionality.

In my case, I'm using React, and I'd like to have more fine-grained control over the panning/zoom when the user's location is found, panning to a viewRect which encompasses both their location and also a target marker for their destination.

Because leaflet-locatecontrol does not expose the native event, I'd need to add a second custom component inside  `<MapContainer>` simply to hook into this event and pass it up to the parent.

KristjanESPERANTO:
Hey folks! Thanks for your input! This has been addressed in a different way via #395 and will be in the next release. 

Instead of adding a callback option, the plugin now fires a `locatelocationfound` event that includes all location data plus a reference to the control instance. This provides the functionality you were looking for (including the one-shot use case) but is more flexible and follows Leaflet's event patterns.

Example:
```js
map.on('locatelocationfound', (e) => {
  console.log(e.latlng, e.accuracy);
  e.control.stop(); // One-shot behavior
});
```
I would appreciate your feedback :slightly_smiling_face:

PR #350, opened 2024-04-09
Export dist module as default instead of src.

RubenNL:
This reduces the bundle size by 60%, 30.85kb to 12.37kb.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.79.0",
   "homepage": "https://github.com/domoritz/leaflet-locatecontrol",
   "description": "A useful control to geolocate the user with many options. Used by osm.org and mapbox among many others.",
-  "main": "src/L.Control.Locate.js",
+  "main": "dist/L.Control.Locate.min.js",
   "author": "Dominik Moritz <domoritz@gmail.com>",
   "repository": {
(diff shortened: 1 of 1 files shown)


domoritz:
Thanks. I did it slightly differently in 453bce6.

RubenNL:
@domoritz I see you changed all of them to the `dist` folder, except the `module` one. Is this a mistake, or is there a reason for this that I can't find?

My project seems to the module export, so v0.81.0 doesn't reduce the project size for me compared to v0.79.0 :(

domoritz:
The module export is for bundlers and the bundler can do its own minimization. It's standard practice I think not to use the minified version for bundlers. Are you running your own minifier?

RubenNL:
I'm using vite, and I think I found the issue. My project is in the middle of a migration to ESM. It seems like "something" is parsing this library as CommonJS, but importing it as ESM. When I manually force the file to be parsed as ESM, it minifies the file correctly.

Thanks for your time, and sorry to bother you :)

EDIT: it was actually using the correct file, but the tool I used to analyze my bunde (vite-bundle-visualizer, built on rollup-bundle-visualizer), reports the size wrong.

PR #357, opened 2024-10-23
Refactor L.Control.Locate source to an esm version of the locate control.

### LINKED ISSUE #313: ES module support?
Hi, thanks for this great plugin. I was wondering if you plan to provide an ES module version of it. The [release announcements of Leaflet 1.9.0](https://github.com/Leaflet/Leaflet/releases/tag/v1.9.0) advocate the use of ES modules and it is planned to even remove the `L` global in the future. It would be great if I could just do

```javascript
import leaflet
import leaflet.locatecontrol
```

in the future :slightly_smiling_face:

prushforth:
Closes #313 

- refactor L.Control.Locate source to esm syntax
- generate umd source from the esm source using grunt-rollup
- include L.Control.Locate.esm.js output in dist
- add demo_esm folder using CDN leaflet via import map
- update URL template to osm tiles in demo directory


demo_esm/index.html
```diff
@@ -0,0 +1,53 @@
...
+        "imports": {
+          "leaflet": "https://cdn.jsdelivr.net/npm/leaflet@1.9.4/dist/leaflet-src.esm.js"
+        }
+      }
+    </script>
+    <link rel="stylesheet" href="../dist/L.Control.Locate.min.css" />
+    <link rel="stylesheet" href="style.css" />
+    <script type="module">
+      import { Map, TileLayer } from 'leaflet';
+      import { LocateControl } from '../dist/L.Control.Locate.esm.js';
```
domoritz: Can we use the src here? Would be nice not to have to compile to test the demo.

domoritz (commented):
Thank you for the refactor and feature.

demo_esm/index.html
prushforth: Yep, done.

prushforth:
Found that leaving creation of window.L in the esm source was problematic for module usage, removed and it's now generated with the rollup footer.  Everything should still work, having restored some inadvertent deletions. Sorry for all the pushes!

yarn.lock
domoritz, line 1: this should be deleted

package-lock.json
domoritz, line 1: We should not mix yarn and this.

domoritz (approved):

mickaeltr:
Hello @prushforth,

I noticed that upgrading from v0.81.1 to v0.82.0 breaks the following code (in a Next.JS 14 application with leaflet 1.9.4):

```ts
import "leaflet.locatecontrol"; // Import plugin
import "leaflet.locatecontrol/dist/L.Control.Locate.min.css"; // Import styles
import L from "leaflet"; // Import L from leaflet to start using the plugin
import "leaflet/dist/leaflet.css";

function createMap(container: HTMLElement): L.Map {
  const map = L.map(container).addLayer(
    L.tileLayer("https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"),
  );
  L.control.locate().addTo(map).start();
  return map;
}
```

Screenshot 2024-11-12 at 12 07 38

I believe that this relates to this PR, so I was wondering if you can spot something to adapt in my code or maybe a bug to fix in leaflet.locatecontrol code or documentation?

Thanks for your help!

domoritz:
@prushforth  can you help?

prushforth:
Since the app is using esm it should use the imported locate() function instead of the L.Control.Locate global, or it could use the new LocateControl({options}) successfully I believe. Note that there's no L global exported by the Leaflet module, but maybe that's created by your build.

domoritz:
Do we need some more docs?

mickaeltr:
I managed to fix this with:
- `import { LocateControl } from "leaflet.locatecontrol";` instead of `import "leaflet.locatecontrol";`
- `new LocateControl()` instead of `L.control.locate()`
- some ESLint tweaks to workaround missing type errors

I would suggest to update:
- the [npm part](https://github.com/domoritz/leaflet-locatecontrol?tab=readme-ov-file#with-npm) of the documentation
- the Typescript [types](https://www.npmjs.com/package/@types/leaflet.locatecontrol)

Many thanks for the help

domoritz:
Added docs in https://github.com/domoritz/leaflet-locatecontrol/commit/b20d77e4184fdfc59ff0037f8d95471a49af6f81

jeffreykog:
The suggested approach:
```ts
import { LocateControl } from "leaflet.locatecontrol";
import "leaflet.locatecontrol/dist/L.Control.Locate.min.css";
```

Does not work currently in a TypeScript project, as the TypeScript definition (https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/leaflet.locatecontrol) still exports the code as being part of the 'leaflet' namespace. This could potentially be changed to support both.

Would it be an option to pull the typescript definitions into this repo and manage it here locally? I could work on a PR for either solution.

domoritz:
I'd be excited about prs. I like typescript.

PR #358, opened 2024-12-05
feat: Add typescript definitions

jeffreykog:
As suggested in https://github.com/domoritz/leaflet-locatecontrol/pull/357#issuecomment-2514216228. After this project was made ESM-ready (very nice!) the suggested way of importing this project is not supported in TypeScript because the TS definitions don't exist in this repo, and the ones from DefinitelyTyped still assume the `L.Control.Locate` global.

Considering the `L.Control.Locate` global only exists in the UMD bundle but not in the ESM bundle, i chose the approach to default to ESM structure in the TS definitions in this repo. `@types/leaflet.locatecontrol` will still exist for the possibility to use the `L.Control.Locate` global in the UMD bundle. This might not be acceptable, so i'm looking for feedback on this. Maybe it is better to support both.


domoritz (changes requested):
Thank you for the pull request and thank you for your patience. I like this and think modeling the esm code makes the most sense since that's what bundlers use and what most people will encounter.

Please add typescript as a dev dependency and add a check to run `npx tsc src/L.Control.Locate.d.ts` (and make sure it succeeds). Right now, I am getting `Cannot find namespace 'L'`.

dist/L.Control.Locate.d.ts
```diff
@@ -0,0 +1,59 @@
...
+  iconElementTag?: string | undefined;
+  textElementTag?: string | undefined;
+  circlePadding?: number[] | undefined;
+  metric?: boolean | undefined;
+  createButtonCallback?: ((container: HTMLDivElement, options: LocateOptions) => { link: HTMLAnchorElement; icon: HTMLElement }) | undefined;
+  onLocationError?: ((event: ErrorEvent, control: LocateControl) => void) | undefined;
+  onLocationOutsideMapBounds?: ((control: LocateControl) => void) | undefined;
+  showPopup?: boolean | undefined;
+  strings?: StringsOptions | undefined;
+  locateOptions?: L.LocateOptions | undefined;
```
domoritz, line 44: This probably needs to be

```suggestion
  locateOptions?: LocateOptions | undefined;
```

domoritz:
Merged manually. Thanks!

PR #359, opened 2024-12-26
Minor changes

KristjanESPERANTO:
This PR contains only a few minor changes. Checkout the commit messages.

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -3,7 +3,6 @@
   "plugins": ["stylelint-prettier"],
   "root": true,
   "rules": {
-    "prettier/prettier": true,
-    "property-no-vendor-prefix": null
+    "prettier/prettier": true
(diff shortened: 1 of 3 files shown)


.stylelintrc.json
```diff
@@ -3,7 +3,6 @@
   "plugins": ["stylelint-prettier"],
   "root": true,
   "rules": {
-    "prettier/prettier": true,
-    "property-no-vendor-prefix": null
```
domoritz, line 7: Why remove this?

domoritz (commented):
Thank you

.stylelintrc.json
KristjanESPERANTO: There are no vendor prefixes in the current code, so I think it doesn't make much sense to deactivate this rule.

PR #364, opened 2025-04-01
Set style as attributes instead of properties

### LINKED ISSUE #363: Use of inline styles on SVG breaks strict Content-Security-Policy (CSP)
The SVG circle does not get its style attribute populated if the CSP does not allow 'unsafe-inline', and turns out black.

To reproduce, serve your website with the Content-Security-Policy header set to something like:

```
Content-Security-Policy: default-src 'self';
```

## Workaround

If you are not using any custom styles, but just the default, you can allow 'unsafe-hashes' and the following hash: `sha256-WKuDjGoTNqYd0JB3SJvtVKJQeOAgCHByrgm2zsb9cD0=`:

```diff
-Content-Security-Policy: default-src 'self';
+Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-hashes' 'sha256-WKuDjGoTNqYd0JB3SJvtVKJQeOAgCHByrgm2zsb9cD0=';
```

The hash represents the following value of `style`, which seems to be the default:

```
<circle r="9" style="stroke:#fff;stroke-width:3;fill:#2A93EE;fill-opacity:1;opacity:1;"></circle>
``` 

## Suggestion

I would suggest using a class for the default styles, and only setting styles via JavaScript if the user provides any custom options.

I guess you could also document the use of 'unsafe-hashes' from the workaround above, but as the names suggests, 'unsafe-hashes' is not the optimal solution.

hlfan:
Fixes #363

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -37,19 +37,19 @@ const LocationMarker = Marker.extend({
     let style = "";
 
     if (opt.color !== undefined) {
-      style += `stroke:${opt.color};`;
+      style += ` stroke="${opt.color}"`;
     }
     if (opt.weight !== undefined) {
-      style += `stroke-width:${opt.weight};`;
+      style += ` stroke-width="${opt.weight}"`;
     }
     if (opt.fillColor !== undefined) {
-      style += `fill:${opt.fillColor};`;
+      style += ` fill="${opt.fillColor}"`;
     }
     if (opt.fillOpacity !== undefined) {
-      style += `fill-opacity:${opt.fillOpacity};`;
+      style += ` fill-opacity="${opt.fillOpacity}"`;
     }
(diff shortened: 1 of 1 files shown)


src/L.Control.Locate.js
```diff
@@ -37,19 +37,19 @@ const LocationMarker = Marker.extend({
     let style = "";
 
     if (opt.color !== undefined) {
-      style += `stroke:${opt.color};`;
+      style += ` stroke="${opt.color}"`;
```
domoritz: Instead of adding the space here, let's add the styles to an array and then join it on `" "`. That makes it easier to guarantee space between items.

domoritz (approved):
Just one code style change.

domoritz (approved):
Nice!

PR #369, opened 2025-05-19
refactor: replace deprecated Leaflet DOM helpers and factory functions for Leaflet 2 compatibility

jeffreykog:
Leaflet v2-alpha was released. A lot of legacy browser support stuff has been removed, and ESM builds are now the default. This PR changes all references to these deprecated/removed classes so leaflet-locatecontrol can be compatible with both v1 and v2 at the same time.

### DIFF
--- a/demo/esm/leaflet2.html
+++ b/demo/esm/leaflet2.html
@@ -1,0 +1,26 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <title>Demo Leaflet.Locate - Leaflet 2</title>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="icon" href="../../favicon.svg" />
+
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/leaflet@2.0.0-alpha.1/dist/leaflet.css" />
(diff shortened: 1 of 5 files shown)


domoritz:
Thanks. I'll wait for the official release before considering this.

simon04:
`Util.requestAnimFrame` has also been removed (line 680).

KristjanESPERANTO:
Since we had a lot of changes since this PR was created, I did a rebase, fixed the merge conflicts and added a demo for Leaflet 2 :slightly_smiling_face:

KristjanESPERANTO:
@domoritz Leaflet v2 has been announced for a while now but the official release keeps slipping - and users are already adopting the alpha to profit from it's improvements.

The changes here are minimal and backwards-compatible with Leaflet v1: one import style change (`setOptions` → `Util.setOptions`) and an updated `peerDependencies`.

I see no risk to existing users and don't think we should let the uncertainty of an upstream release date hold back users who want to adopt Leaflet 2 today.

Would you be open to merging this now? :slightly_smiling_face:

domoritz:
Makes sense. Let's merge.

KristjanESPERANTO:
Great! Are you going to make a new release, or should I? :slightly_smiling_face:

PR #371, opened 2025-07-09
Fix: fallback for  missing orientation data on IOS Safari

### LINKED ISSUE #370: Compass needle always points north on iOS Safari
I get 2 permission prompts, one for location and one for orientation. I accept them both, the compass needle is drawn but no matter how I move my phone the needle always points north

https://github.com/openstreetmap/openstreetmap-website/issues/6030

On macOS Safari it correctly never prompts for the orientation and doesn't draw the compass needle.

LaurentSauzetOff:
🧭 Context

Fixes #370 – On iOS Safari, the compass needle always points north because no valid orientation data is received, even after permission is granted.
✅ What this does

    Adds a timeout-based fallback to detect if no orientation data is received

    Automatically hides the compass needle after 2 seconds if alpha remains null

    Adds a new internal method _hideCompass() for better handling of compass visibility

🧪 Tested on:

    iOS 17.5 Safari ✅ (bug previously present, now handled gracefully)

    Android Chrome ✅

    Desktop Chrome/Firefox ✅

No regressions noticed on non-iOS platforms.

Let me know if you'd like me to squash the commits or rebase it.


eslint.config.js
```diff
@@ -13,6 +13,23 @@ module.exports = [
       }
     }
   },
+  {
+    files: ["src/L.Control.Locate.js"],
+    languageOptions: {
```
domoritz, line 18: Let's use English.

src/L.Control.Locate.js
```diff
@@ -22,7 +22,7 @@ const removeClasses = (el, names) => {
  * Compatible with Circle but a true marker instead of a path
  */
 const LocationMarker = Marker.extend({
-  initialize(latlng, options) {
+  initialize: function(latlng, options) {
```
domoritz, line 25: Let's not change unrelated things here.

src/L.Control.Locate.js
```diff
@@ -467,30 +467,53 @@ const LocateControl = Control.extend({
...
-        if (DeviceOrientationEvent && typeof DeviceOrientationEvent.requestPermission === "function") {
-          DeviceOrientationEvent.requestPermission().then(function (permissionState) {
-            if (permissionState === "granted") {
-              deviceorientation();
-            }
-          });
+
+        window.addEventListener(eventName, handler, { once: true });
+
+        // Fallback après 2 secondes si rien n’est reçu
```
domoritz, line 486: Same as above. Use English.

domoritz (commented):
Thanks for the pull request! A few comments.

domoritz:
Needs to be redone with comments addressed.

PR #374, opened 2025-09-05
Fix rotation of compass arrow on Safari and Firefox

### LINKED ISSUE #370: Compass needle always points north on iOS Safari

draperunner:
Using transform on the `<svg>` element is not supported by Firefox or Safari: https://caniuse.com/mdn-svg_global_attributes_transform_svg_root

Therefore, this has not worked since PR #368 / commit 4cd1349a738ee2986e6341e05b846b6e491cc585.

Using style/CSS works though, as it was before PR #364 / commit e2d60a58b39ebffff86ec1378b4c374caf1016bd

I tried some workarounds on the svg by adding a `<g>` element with rotation, but I did not get it to work as expected. I gave up and went back to using CSS.

Since the point of #364 was to avoid inline styles to allow strict Content Security Policies, I found a way to update the style without using the `style` attribute, but instead using a class name and editing the style associated with it. This works, and you can use this with style-src 'self' in the CSP.

Fixes #370

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -88,11 +88,17 @@ const CompassMarker = LocationMarker.extend({
     setOptions(this, options);
     this._latlng = latlng;
     this._heading = heading;
+    this._arrowClass = "leaflet-control-locate-heading-arrow";
+    this._cssRule = this._getSvgCSSRule();
     this.createIcon();
   },
 
   setHeading(heading) {
     this._heading = heading;
+
+    if (this._cssRule) {
+      this._cssRule.style.setProperty("transform", `rotate(${this._heading}deg)`);
+    }
   },
 
   /**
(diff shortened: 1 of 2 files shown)


domoritz (commented):
Thanks for the pull request. Is there a simpler way to do this? This feels quite cumbersome.

draperunner:
@domoritz I had a go at calculating the correct, "rotated", path for the compass triangle without having to use CSS transforms in the last commit. Maybe less cumbersome? I find it good that we don't have to look up and manipulate style sheet rules.

PR #381, opened 2025-12-28
Demo improvements

KristjanESPERANTO:
I put a little more effort into the demos and the landing page :slightly_smiling_face: 

- Replace CARTO dark tiles (unclear licensing) with CSS filter approach (clear license situation, looks better, faster loading)
- Reorganize demos into `demo/{umd,esm,mapbox}/` subdirectories
- Improve landing page with gradient header, "logo", links, and better descriptions
- Add favicon with locate control icon
- Share common CSS across all demos
- Add proper DOCTYPE and lang attributes

You can see the result here: https://kristjanesperanto.github.io/leaflet-locatecontrol/

To compare, the current status here: https://domoritz.github.io/leaflet-locatecontrol/


domoritz:
Can you least align the text at the bottom? Makes it more readable.

KristjanESPERANTO:
Do you mean align to the left? Something like this?

image

To compare - this is the current state of the PR:
image

This is the current state in the repo:
image

I also considered whether it might be better to put it in a header instead of a footer. Visually, it wouldn't look as nice as a footer, but it might be more practical. What do you think?

Draft 1:
image

Or, even simpler, just below the text. That's where users will find it most easily. That feels like my new favorite.

image

What do you think? :slightly_smiling_face:

domoritz:
Below text is good

KristjanESPERANTO:
Done incl. adding the "logo" next to the title and removing the numbering from the demo buttons :slightly_smiling_face: 

image

KristjanESPERANTO:
Thanks for merging!

@domoritz Oh, after changing the demo directory, you have to change the URL for the repository homepage. This can not done via a PR.

It's now: https://domoritz.github.io/leaflet-locatecontrol/demo/

But should be: https://domoritz.github.io/leaflet-locatecontrol/

image

PR #385, opened 2026-01-05
feat: stop locate when control is removed from map

yohanboniface:
If there is an active location when the control is removed from the map, let's call stop to deactivate it, and remove event listeners (zoomEnd…).

Thanks!

### DIFF
--- a/src/L.Control.Locate.d.ts
+++ b/src/L.Control.Locate.d.ts
@@ -56,6 +56,8 @@ export class LocateControl extends Control {
 
   onAdd(map: Map): HTMLElement;
 
+  onRemove(): void;
+
   start(): void;
 
   stop(): void;
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -380,6 +380,13 @@ const LocateControl = Control.extend({
   },
 
   /**
+   * Called when control is removed from the map.
+   */
+  onRemove() {
+    this.stop()
+  },
+
+  /**
    * This method is called when the user clicks on the control.
    */
   _onClick() {


domoritz (approved):
Thanks! Please fix the lint issue so we can merge.

KristjanESPERANTO:
@yohanboniface No pressure, but are you working on this? :slightly_smiling_face:

yohanboniface:
@KristjanESPERANTO sorry, I've updated the PR now :)

src/L.Control.Locate.d.ts
```diff
@@ -56,6 +56,8 @@ export class LocateControl extends Control {
 
   onAdd(map: Map): HTMLElement;
 
+  onRemove(map: Map): void;
```
KristjanESPERANTO: Then map should be removed here too, right?

yohanboniface: Oops! Done :)

KristjanESPERANTO (approved):

src/L.Control.Locate.d.ts
KristjanESPERANTO: Thanks! 😃

domoritz (changes requested):
Lint is still failing.

yohanboniface:
Nevermind!

domoritz:
Would you like help with this pull request? If you still think it's a good fix, we can help to get it in.

src/L.Control.Locate.js
```diff
@@ -386,6 +379,13 @@ const LocateControl = Control.extend({
     return container;
   },
 
+  /**
+   * Called when control is removed from the map.
+   */
+  onRemove() {
+    this.stop()
```
KristjanESPERANTO, line 386: In this line the ; was missing.

KristjanESPERANTO:
I don't understand why the PR was cancelled, given that the change made sense. So, I created a new PR for this change: #388.

In my opinion, using a linter for PR quality control is best practice. It's a pity that this disrupted the flow here.

domoritz:
Agreed. I set up this repo a long time ago before easy linting was available. I'd be happy to add something to make it less disruptive for others to contribute. I love when others contribute to packages!

PR #386, opened 2026-01-08
feat: Add aria label to button

robinroloff:


domoritz:
Doesn't the title already expose the information?

robinroloff:
> Doesn't the title already expose the information?

We got a warning from semrush that the links do not have a valid anchor text. The "title" attribute is not a reliable solution for screenreaders.

domoritz:
I don't know what semrush is. Can you point to WCAG guidelines or test in a screen reader to confirm that this is making things better?

robinroloff:
Semrush is a SEO-Scanner tool. 
[On this page](https://www.w3.org/WAI/WCAG22/Understanding/link-purpose-link-only) in the section "Sufficient Techniques" you can see that having a title attribute is not listed there.

domoritz:
Thanks for digging that up. Makes sense.

robinroloff:
Thanks for merging. Do you think you can get a release going for this soon?

domoritz:
leaflet.locatecontrol@0.86.0 is out

PR #388, opened 2026-01-14
feat: stop locate when control is removed from map

KristjanESPERANTO:
If there is an active location when the control is removed from the map, let's call stop to deactivate it, and remove event listeners (zoomEnd…).

Original author of this change: @yohanboniface.

I think this is a useful change, so I picked up PR #385 and fixed the linting issue that caused the PR to fail.

### DIFF
--- a/src/L.Control.Locate.d.ts
+++ b/src/L.Control.Locate.d.ts
@@ -56,6 +56,8 @@ export class LocateControl extends Control {
 
   onAdd(map: Map): HTMLElement;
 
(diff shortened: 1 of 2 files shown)


domoritz:
Thank you and thanks for including @yohanboniface in the commit so they get the credit as well.

domoritz (approved):

KristjanESPERANTO:
Okay, let's go with it :slightly_smiling_face:

PR #390, opened 2026-01-24
test: add test framework

KristjanESPERANTO:
This PR adds a test framework.

- **41 unit tests** using lightweight test setup (`node:test` + `happy-dom`)
- **CI now runs** lint → test → build
- **`preversion` hook** prevents releases if tests fail

Since I want to tackle other issues and do some refactoring, the tests will be extremely valuable.

---

**Note:** Initially, I had a fix commit here in this PR. After thinking about it, I moved it to a separate PR (#391) to keep the git history clean.

Since the fix is no longer included in this PR, the tests are failing, as expected.


src/L.Control.Locate.js
```diff
@@ -335,20 +335,38 @@ const LocateControl = Control.extend({
...
-    for (const i in options) {
-      if (typeof this.options[i] === "object") {
-        extend(this.options[i], options[i]);
+  initialize(options = {}) {
+    // Clone default options to prevent prototype pollution.
+    // Without this, all instances share the same nested objects
+    // (e.g., strings, clickBehavior) and modifying one affects all.
+    const defaults = this.options;
+    this.options = {};
+    for (const key in defaults) {
```
domoritz: Can we just do a structured clone?

src/L.Control.Locate.js
```diff
@@ -335,20 +335,38 @@ const LocateControl = Control.extend({
...
+        this.options[key] = [...val];
+      } else if (val?.constructor === Object) {
+        this.options[key] = { ...val };
       } else {
-        this.options[i] = options[i];
+        this.options[key] = val;
       }
     }
 
-    // extend the follow marker style and circle from the normal style
```
domoritz, line 348: I think we can keep the simpler extend if we use structured clone.

domoritz (changes requested):

src/L.Control.Locate.js
KristjanESPERANTO: I agree the current solution feels over-engineered. I tried using `structuredClone`, but it throws on functions and class references (we have `createButtonCallback`, `onLocationError`, `markerClass`, `compassClass`, etc. in the options).

The generic shallow-clone approach works, but I'm not so happy with it either. Do you have a simpler pattern in mind that handles the function/class references?

domoritz: Maybe pulling out the clone into a separate method makes it cleaner?

src/L.Control.Locate.js
KristjanESPERANTO: Good suggestion! I've extracted the clone logic into `_cloneOptions()`:

- `initialize()` is now shorter and more readable
- The clone logic is isolated and documented
- JSDoc comment explains the "why" (preventing prototype pollution)

Updated in the latest commit.

Ah, should I perhaps split the PR into two: one for adding the test setup and one for the fix? For a cleaner git history (since we are using squash here).

src/L.Control.Locate.js
```diff
@@ -335,20 +335,44 @@ const LocateControl = Control.extend({
...
-  initialize(options) {
-    // set default options if nothing is set (merge one step deep)
-    for (const i in options) {
-      if (typeof this.options[i] === "object") {
-        extend(this.options[i], options[i]);
+  /**
+   * Shallow clone options to prevent prototype pollution.
+   * Clones arrays and plain objects, keeps functions/classes as references.
+   */
+  _cloneOptions(options) {
```
domoritz: Instead of putting this on the object, let's move it to the top level (similarly to add and remove classes). We probably should make all three normal functions.

src/L.Control.Locate.js
domoritz: Splitting is nice but I am okay with not making you extra work.

src/L.Control.Locate.js
KristjanESPERANTO: Done :slightly_smiling_face:

src/L.Control.Locate.js
KristjanESPERANTO: I'll do the extra work. It's not too much effort :slightly_smiling_face:

PR #391, opened 2026-01-25
fix: prevent prototype pollution in LocateControl options

KristjanESPERANTO:
Found a subtle prototype pollution issue: nested options like `strings`, `clickBehavior`, and style objects were shared between all instances. If you customized options on one control, it would leak into others:

```javascript
const control1 = new LocateControl({ strings: { title: "Find me" } });
const control2 = new LocateControl();
// Oops: control2 also had title "Find me"
```

Fixed by shallow-cloning all options in `initialize()` so each instance gets its own copy.

This bug was rare in practice (most apps only have one LocateControl), but it could bite you in SPAs or dashboards with multiple maps.

---
**Note:** This commit was previously in PR #390. To keep the git history clean, I put this commit in this separate PR.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -5,7 +5,7 @@ This file is part of the leaflet locate control. It is licensed under the MIT li
 You can find the project at: https://github.com/domoritz/leaflet-locatecontrol
 */
 
-import { Control, Marker, DomUtil, setOptions, divIcon, extend, LayerGroup, circle, DomEvent, Util as LeafletUtil } from "leaflet";
+import { Control, Marker, DomUtil, setOptions, divIcon, LayerGroup, circle, DomEvent, Util as LeafletUtil } from "leaflet";
 const addClasses = (el, names) => {
   names.split(" ").forEach((className) => {
     el.classList.add(className);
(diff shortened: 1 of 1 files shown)


src/L.Control.Locate.js
```diff
@@ -18,6 +18,25 @@ const removeClasses = (el, names) => {
   });
 };
 
+/**
+ * Shallow clone options to prevent prototype pollution.
+ * Clones arrays and plain objects, keeps functions/classes as references.
+ */
+const _cloneOptions = (options) => {
```
```suggestion
const cloneOptions = (options) => {
```

domoritz: And probably could just be a function instead of an arrow function.

domoritz (approved):

src/L.Control.Locate.js
KristjanESPERANTO: Yes, both done :slightly_smiling_face: 

After this PR I will create a refactor PR and would, among other things, change `removeClasses` and `addClasses` from arrow functions to regular functions.

src/L.Control.Locate.js
```diff
@@ -335,20 +356,25 @@ const LocateControl = Control.extend({
...
+        this.options[key] = userVal;
       }
     }
 
-    // extend the follow marker style and circle from the normal style
-    this.options.followMarkerStyle = extend({}, this.options.markerStyle, this.options.followMarkerStyle);
-    this.options.followCircleStyle = extend({}, this.options.circleStyle, this.options.followCircleStyle);
-    this.options.followCompassStyle = extend({}, this.options.compassStyle, this.options.followCompassStyle);
+    // Follow styles inherit from base styles
+    Object.assign(this.options.followMarkerStyle, this.options.markerStyle, this.options.followMarkerStyle);
```
domoritz, line 375: How does this differ from https://leafletjs.com/reference.html#util-extend?

domoritz (approved):

src/L.Control.Locate.js
KristjanESPERANTO: For this use case, there's no functional difference — both copy properties from source objects to the target.

`Object.assign` is the native JavaScript equivalent. I chose it because `L.Util.extend` will be removed in Leaflet 2 (see [alpha release notes](https://leafletjs.com/2025/05/18/leaflet-2.0.0-alpha.html)), so this avoids a future migration step.

Or what do you think?

domoritz: Sounds good since it's going away.

PR #392, opened 2026-01-27
refactor: modernize code and improve maintainability

KristjanESPERANTO:
This PR refactors the codebase by replacing deprecated APIs, simplifying code patterns, and fixing a potential memory leak. These changes are also a step to Leaflet 2 compatibility.

See commit messages for details. If that's too many changes for one PR, let me know :slightly_smiling_face: 

All existing tests pass. Added 4 new tests for popup binding scenarios.

### DIFF
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -755,6 +755,7 @@ const LocateControl = Control.extend({
     this._layer.clearLayers();
     this._marker = undefined;
     this._circle = undefined;
+    this._compass = undefined;
   },
 
   /**
@@ -944,6 +945,8 @@ const LocateControl = Control.extend({
     } else if (this.options.setView === "untilPanOrZoom") {
       return !this._userPanned && !this._userZoomed;
     }
+
+    return false;
   },
 
   /**
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -638,10 +638,10 @@ const LocateControl = Control.extend({
           padding: this.options.circlePadding,
           maxZoom: this.options.initialZoomLevel || this.options.locateOptions.maxZoom
         });
-        LeafletUtil.requestAnimFrame(function () {
+        requestAnimationFrame(() => {
           // Wait until after the next animFrame because the flyTo can be async
           this._ignoreEvent = false;
-        }, this);
+        });
       }
     }
   },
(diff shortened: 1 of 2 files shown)


src/L.Control.Locate.js
```diff
@@ -450,8 +463,8 @@ const LocateControl = Control.extend({
         case "stop":
           this.stop();
           if (this.options.returnToPrevBounds) {
-            const f = this.options.flyTo ? this._map.flyToBounds : this._map.fitBounds;
-            f.bind(this._map)(this._prevBounds);
+            const method = this.options.flyTo ? "flyToBounds" : "fitBounds";
+            this._map[method](this._prevBounds);
```
domoritz: Doesn't this break with strong minification?

src/L.Control.Locate.js
```diff
@@ -623,26 +623,29 @@ const LocateControl = Control.extend({
     if (this._isOutsideMapBounds()) {
       this._event = undefined; // clear the current location so we can get back into the bounds
       this.options.onLocationOutsideMapBounds(this);
+      return;
+    }
+
+    const latlng = this._event.latlng;
+
+    if (this._justClicked && this.options.initialZoomLevel !== false) {
+      const method = this.options.flyTo ? "flyTo" : "setView";
```
domoritz: same as above

src/L.Control.Locate.js
```diff
@@ -623,26 +623,29 @@ const LocateControl = Control.extend({
...
-        f.bind(this._map)(this.options.getLocationBounds(this._event), {
-          padding: this.options.circlePadding,
-          maxZoom: this.options.initialZoomLevel || this.options.locateOptions.maxZoom
-        });
-        requestAnimationFrame(() => {
-          // Wait until after the next animFrame because the flyTo can be async
-          this._ignoreEvent = false;
-        });
-      }
+      const method = this.options.flyTo ? "flyToBounds" : "fitBounds";
```
domoritz: same

domoritz (approved):
Let's remove 98f82854a4e7678859f10fd82922364ba00dceea since it makes the code harder to reason about. I don't like calling methods by string name if there is a cleaner way to do it.

src/L.Control.Locate.js
KristjanESPERANTO: I don't think so. The minifier only renames properties within the bundled code itself, not references to external library APIs.

KristjanESPERANTO:
> Let's remove https://github.com/domoritz/leaflet-locatecontrol/commit/98f82854a4e7678859f10fd82922364ba00dceea since it makes the code harder to reason about. I don't like calling methods by string name if there is a cleaner way to do it.

I see what you mean. No problem. I just threw out the commit :slightly_smiling_face:

src/L.Control.Locate.js
KristjanESPERANTO: After the removal of 98f82854a4e7678859f10fd82922364ba00dceea it's no longer relevant anyway.

PR #395, opened 2026-01-28
feat: add locatelocationfound event

KristjanESPERANTO:
I looked at PR #343, which has been open since 2023. The implementation had issues, but the use case is valid. This PR here takes a different approach - using an event instead of a callback option - which is more flexible and consistent with Leaflet's patterns.

## Why is this useful?

Some practical scenarios I can imagine would need access to the control when location is found:

- **One-shot location** - Get user location once, then stop (e.g., "show nearby restaurants")
- **Custom zoom** - Fit bounds to include user + target marker (urb4nc4rl's use case from #343)
- **Geofencing** - Check if user is in delivery area
- **Analytics** - Track accuracy and behavior
- **Form autofill** - Reverse geocode address

## Why add it to the plugin?

**"Can't you just use `map.locate()` directly without the plugin?"**

Yes, but the plugin provides UI, marker rendering, accuracy circles, compass, zoom behavior, popups, error handling, and visual states. Most apps use the plugin for this functionality.

**"Then just listen to the native `locationfound` event?"**

You can:

```js
const lc = L.control.locate().addTo(map);
map.on('locationfound', (e) => {
  lc.stop();
});
```

But this requires keeping a separate control reference - extra boilerplate.

## What this PR does

Adds a `locatelocationfound` event with direct control access:

```js
map.on('locatelocationfound', (e) => {
  console.log(e.latlng, e.accuracy);
  e.control.stop(); // No separate reference needed
});
```

This matches the existing pattern - `locateactivate`, `locatedeactivate`, and `locationtimeout` already pass the control instance.

**Changes:**
- New event with TypeScript types
- Tests added
- Docs updated

### DIFF
--- a/src/L.Control.Locate.d.ts
+++ b/src/L.Control.Locate.d.ts
@@ -1,6 +1,16 @@
 import { Control, Layer, Map, ControlOptions, PathOptions, MarkerOptions, LocationEvent, LatLngBounds, LocateOptions as LeafletLocateOptions } from "leaflet";
 
 export type SetView = false | "once" | "always" | "untilPan" | "untilPanOrZoom";
+
+/**
+ * Event fired when a location is found by the locate control.
+ * Extends Leaflet's LocationEvent with a reference to the control instance.
+ */
+export interface LocateLocationFoundEvent extends LocationEvent {
+  /** Reference to the locate control instance */
+  control: LocateControl;
+}
+
 export type ClickBehavior = "stop" | "setView";
 
 export interface StringsOptions {
--- a/src/L.Control.Locate.js
+++ b/src/L.Control.Locate.js
@@ -882,6 +882,12 @@ const LocateControl = Control.extend({
     this._drawMarker();
     this._updateContainerStyle();
 
+    // Fire event with location data and control reference
+    this._map.fire("locatelocationfound", {
+      ...e,
+      control: this
+    });
+
     switch (this.options.setView) {
       case "once":
         if (this._justClicked) {
--- a/README.md
+++ b/README.md
@@ -194,9 +194,26 @@ You can keep the plugin active but stop following using `lc.stopFollowing()`.
 
 You can leverage the native Leaflet events `locationfound` and `locationerror` to handle when geolocation is successful or produces an error. You can find out more about these events in the [Leaflet documentation](https://leafletjs.com/examples/mobile/#geolocation).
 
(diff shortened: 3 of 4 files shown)


src/L.Control.Locate.js
```diff
@@ -882,6 +882,19 @@ const LocateControl = Control.extend({
...
+      latlng: e.latlng,
+      accuracy: e.accuracy,
+      altitude: e.altitude,
+      altitudeAccuracy: e.altitudeAccuracy,
+      heading: e.heading,
+      speed: e.speed,
+      timestamp: e.timestamp,
+      bounds: e.bounds,
+      control: this
+    });
```
domoritz, line 889: Would this work?

```suggestion
    // Fire event with location data and control reference
    this._map.fire("locatelocationfound", {
      ...e,
      control: this
    });
```

src/L.Control.Locate.d.ts
```diff
@@ -1,6 +1,41 @@
...
+  LocateOptions as LeafletLocateOptions,
+  LatLng
+} from "leaflet";
 
 export type SetView = false | "once" | "always" | "untilPan" | "untilPanOrZoom";
+
+/**
+ * Event fired when a location is found by the locate control.
+ */
+export interface LocateLocationFoundEvent {
```
domoritz: Is this type somewhere in leaflet as well? Would be nice not to redeclare.

domoritz (approved):

src/L.Control.Locate.js
KristjanESPERANTO: Wow. That's much more elegant and future-proof - if Leaflet ever adds new properties to the locationfound event, they will be automatically passed through. Just changed it.

src/L.Control.Locate.d.ts
KristjanESPERANTO: No, `LocateLocationFoundEvent` doesn't exist in Leaflet. But I just changed it to `extends LocationEvent` - so we only declare what's actually new (the `control` property) and inherit everything else from Leaflet's existing type. Much cleaner, just like the `...e` spread in the implementation.

domoritz (approved):

KristjanESPERANTO:
Just removed the `LatLng` import after your approval - not needed since the switch to `extends LocationEvent` -  `LocationEvent` already includes it.

Issue #334, opened 2023-04-08
Not finding location in Chrome & Firefox for Mobile


walkbmc:
Locator is not finding a location on either Chrome or Firefox for mobile, just spins and spins, no errors or anything

nitinkg7:
I am also facing this issue in ios webview

Mary-Tyler-Moore:
I'm facing this same issue. The demo website ( https://domoritz.github.io/leaflet-locatecontrol/demo/ ) works on Android when using Chrome, and work on iOS when using Safari, but does not work on iOS when using Chrome.

I've noticed that when it works two messages are displayed:
1: "domoritz.github.io" would like to use your current location.
2: "domoritz.github.io" Would like to Access Motion and Orientation

![iOS_Safari_1](https://github.com/domoritz/leaflet-locatecontrol/assets/46167401/5fb23446-7cad-436f-8bbc-cfa9ea54ac47)

![iOS_Safari_2](https://github.com/domoritz/leaflet-locatecontrol/assets/46167401/12b43814-a559-4ddd-af95-31630edc528a)

And when it doesn't work, when using the Chrome browser on an iOS device, only one message is displayed:
1: "domoritz.github.io" Would like to Access Motion and Orientation

![iOS_Chrome_1](https://github.com/domoritz/leaflet-locatecontrol/assets/46167401/17c5d0c1-5d0a-408b-b4df-3b9dcf7b2103)

![iOS_Chrome_2](https://github.com/domoritz/leaflet-locatecontrol/assets/46167401/fc228c68-664c-415e-93d7-20e8905b8e79)

guillaume-coulon:
Same issue here. Have you figured this out ?

Mary-Tyler-Moore:
@tchercheuse - I decided to create a custom Leaflet control that allows users to locate their position on the map using the browser's geolocation capabilities.

`
L.Control.UserLocation = L.Control.extend({
    options: {
        position: 'bottomright',  // Changed position to 'bottomright'
        title: 'Locate user'
    },

    onAdd: function (map) {
        const controlDiv = L.DomUtil.create('div', 'leaflet-bar leaflet-control leaflet-control-userlocation');
        L.DomEvent
            .addListener(controlDiv, 'click', L.DomEvent.stopPropagation)
            .addListener(controlDiv, 'click', L.DomEvent.preventDefault)
            .addListener(controlDiv, 'click', () => this._locateUser(map));

[...]

PR #401, opened 2026-02-27
refactor: improve compass activation/deactivation and error handling

KristjanESPERANTO:
On iOS Chrome and other WKWebView-based browsers, `DeviceOrientationEvent.requestPermission()` can reject, causing an unhandled promise rejection in the console. This PR adds proper error handling and refactors the compass logic into dedicated methods.

Note: The underlying geolocation issue on iOS Chrome (#334) is a platform limitation that cannot be fixed in JavaScript. It may no longer occur on newer iOS versions — no new reports have been filed since 2023.

### Error handling

- Wrap `DeviceOrientationEvent.requestPermission()` in `try/catch` so permission errors are caught and logged as a warning instead of producing unhandled rejections.

### Refactoring

- Extract `_activateCompass()` — compass setup (feature detection, permission request, event binding) in its own `async` method. Removes `const _this = this`, nested function declaration, and deep nesting.
- Extract `_deactivateCompass()` — symmetric counterpart for cleanup.
- Store bound event name in `_compassEventName` instead of re-running feature detection in `_deactivate()`.

### Tests

- 12 new tests covering compass activation and deactivation (permission granted/denied/rejected, feature detection, event binding/unbinding, cleanup).

## Breaking changes

None.

### DIFF
--- a/spec/L.Control.Locate.spec.js
+++ b/spec/L.Control.Locate.spec.js
@@ -1,4 +1,4 @@
-import { describe, it, beforeEach, afterEach } from "node:test";
+import { describe, it, beforeEach, afterEach, mock } from "node:test";
 import assert from "node:assert";
 import "./setup.js";
 import { Map, LayerGroup } from "leaflet";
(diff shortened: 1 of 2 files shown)


KristjanESPERANTO:
@domoritz I consider this a minor change, but I'm not sure whether I should merge it without your review.

domoritz (approved):

PR #404, opened 2026-03-03
refactor: replace Leaflet DOM helpers with native equivalents

KristjanESPERANTO:
Small refactor PR that swaps out a few Leaflet utility calls for their native browser equivalents. No behaviour changes. This is also a further step towards Leaflet 2 compatibility, as these helpers are removed in Leaflet 2.

- `DomUtil.removeClass` → `removeClasses()` (module-local helper that already existed in the file)
- `divIcon()` factory function → `new DivIcon()` constructor
- `DomUtil.create()` (×4) → a small `createElement()` helper wrapping `document.createElement`. `DomUtil` is no longer imported at all.

Also adds tests for the CSS state transitions (`requesting`, `active`, `following`) that were previously untested.

### DIFF
--- a/spec/L.Control.Locate.spec.js
+++ b/spec/L.Control.Locate.spec.js
@@ -363,6 +363,62 @@ describe("LocateControl", () => {
     });
   });
 
+  describe("CSS state", () => {
+    it("should remove state classes when stopped", () => {
+      const control = new LocateControl();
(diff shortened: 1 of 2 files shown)


KristjanESPERANTO:
Since I consider this PR to be a minor change, I'm merging it myself.

PR #405, opened 2026-03-04
feat: add speed and heading to popup template data

### LINKED ISSUE #345: Request: Add speed and track information
Wouldn't it be amazing to have speed and track information available on the popup?

KristjanESPERANTO:
With this the following placeholders are now available in `strings.popup`:

- `{speed}` – raw value in m/s as provided by the Geolocation API, "N/A" if unavailable
- `{heading}` – degrees 0–360, "N/A" if unavailable

Unit conversion is left to the consumer, e.g. via the function form of `strings.popup`. The default popup template is unchanged.

signal-2026-03-04-220739_002

*The screenshot is no longer up to date. We now display m/s instead of km/h.*

It resolves #345 and is a good addition to PR #398.

Demo: https://kristjanesperanto.github.io/leaflet-locatecontrol/demo/esm/
*Use a mobile and the second locate control at the bottom*.

### DIFF
--- a/demo/esm/script.js
+++ b/demo/esm/script.js
@@ -48,12 +48,12 @@ new LocateControl({
   }
 }).addTo(map);
 
-// Control with text (bottom left) - demonstrates lat/lng/altitude in popup
+// Control with text (bottom left) - demonstrates all popup fields incl. speed & heading
 new LocateControl({
   position: "bottomleft",
   strings: {
     title: "Show me where I am, yo!",
     text: "Locate me",
-    popup: "Lat: {lat}<br>Lng: {lng}<br>Alt: {altitude} {unit}<br>Accuracy: ±{distance} {unit}"
+    popup: "{lat}, {lng}<br>Accuracy: ±{distance} {unit}<br>Altitude: {altitude} {unit}<br>Speed: {speed} {speedUnit}<br>Heading: {heading}°"
   }
 }).addTo(map);
--- a/demo/esm/script.js
+++ b/demo/esm/script.js
@@ -54,6 +54,6 @@ new LocateControl({
   strings: {
     title: "Show me where I am, yo!",
     text: "Locate me",
-    popup: "{lat}, {lng}<br>Accuracy: ±{distance} {unit}<br>Altitude: {altitude} {unit}<br>Speed: {speed} {speedUnit}<br>Heading: {heading}°"
+    popup: "{lat}, {lng}<br>Accuracy: ±{distance} {unit}<br>Altitude: {altitude} {unit}<br>Speed: {speed} m/s<br>Heading: {heading}°"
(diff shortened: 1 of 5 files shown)


demo/esm/script.js
```diff
@@ -48,12 +48,12 @@ new LocateControl({
...
 
-// Control with text (bottom left) - demonstrates lat/lng/altitude in popup
+// Control with text (bottom left) - demonstrates all popup fields incl. speed & heading
 new LocateControl({
   position: "bottomleft",
   strings: {
     title: "Show me where I am, yo!",
     text: "Locate me",
-    popup: "Lat: {lat}<br>Lng: {lng}<br>Alt: {altitude} {unit}<br>Accuracy: ±{distance} {unit}"
+    popup: ["{lat}, {lng}", "Accuracy: ±{distance} {unit}", "Altitude: {altitude} {unit}", "Speed: {speed} {speedUnit}", "Heading: {heading}°"].join("<br>")
```
domoritz: Let's not make this more complicate with a join.

src/L.Control.Locate.d.ts
```diff
@@ -18,6 +18,8 @@ export interface StringsOptions {
   text?: string | undefined;
   metersUnit?: string | undefined;
   feetUnit?: string | undefined;
+  kmhUnit?: string | undefined;
+  mphUnit?: string | undefined;
```
domoritz: what about m/s, knots, etc? I feel we might want to revisit this entirely to have users do the conversion (and we just provide metric information). Thoughts?

domoritz (changes requested):

src/L.Control.Locate.d.ts
KristjanESPERANTO: Good point! I initially followed the pattern of the other string options without thinking it through. Your knots example illustrates it well - and it goes further: you might also want to switch units dynamically based on the value (e.g. m/s when slow, km/h when moving). There's no obvious "right" choice, and trying to cover all cases on our end would get messy fast.

That said, there's an argument for handling at least the common conversions centrally, so every consumer doesn't have to reimplement them. I'm not sure where the right line is - could easily tip into overengineering.

For now I've changed it to exposing the raw m/s value from the Geolocation API and leaving conversion to the consumer. Happy to keep discussing if you have a different direction in mind!

demo/esm/script.js
KristjanESPERANTO: Switched to string :slightly_smiling_face:

src/L.Control.Locate.js
domoritz, line 829: This method is just too limited. We do support passing in a function that gets access to the data. I think what we do now makes sense.

domoritz (approved):

PR #418, opened 2026-05-01
chore: update devDependencies, fix Android compass heading

KristjanESPERANTO:
What started as a routine dependency bump turned into a small bug hunt.

After updating `happy-dom`, four compass tests went red. Turned out the new version always reports `deviceorientationabsolute` as present in `window`, so `_activateCompass` now registers on `deviceorientationabsolute` instead of `deviceorientation`. The tests were dispatching the hardcoded string and never reached the handler.
Fix: use `control._compassEventName` - whatever was actually registered.

While digging into that, I noticed the handler condition was also wrong:

```js
// before
} else if (e.absolute && e.alpha) {
// after
} else if (e.alpha !== null) {
```

The handler is registered for exactly one event type, so checking `e.absolute` is redundant regardless of what value it carries. On top of that, `alpha === 0` (due North) would have been silently ignored too.

The dep bump didn't cause a bug - it just made one visible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,17 +9,17 @@
       "version": "0.89.0",
       "license": "MIT",
       "devDependencies": {
-        "@eslint/css": "^0.14.1",
+        "@eslint/css": "^1.2.0",
         "@eslint/js": "^10.0.1",
         "@rollup/plugin-terser": "^1.0.0",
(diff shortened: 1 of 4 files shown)


domoritz (approved):
Good catch!

